PR #280, opened 2015-08-20
V0.4.2 318 extract search

fredkingham:
fixed the spacing so that we don't overlap.
we now use bootstrap tooltips and a new focus directive to drive the user through the flow.
saved searches are encapsulated on the right.

What's not fixed is some of the saved search naming stuff which should probably be part of a separate ticket.


fredkingham:
@gabpoll would you mind taking a look at this on the heroku instance? Its the advanced search, feature, David sent through a preview yesterday. I had to tweak a few things and I've changed some of the structure.

I think its intuitive but that could be because I made it :), let me know what you think

GabPoll:
looks great @fredkingham 

not 100% sure what this particular ticket refers to, so keeping open.

The clear search button doesn't quite work - clears first line of a 2 stage search, but second line then becomes 1st line of a new search. Is there a separate ticket for that?

fredkingham:
@GabPoll no that's a bug, only happens on some occurrences, good spot, fixing now

GabPoll:
this is closed - do you need a new ticket for the search bug i picked up?

fredkingham:
nope, I closed this branch because it should have been against a version not master, (still getting the hang of things...)

I've fixed the bug and submitted a new pull request https://github.com/openhealthcare/opal/tree/v0.4.2-318-extract-search

it *should* now be fixed on Heroku

GabPoll:
it is fixed indeed.

Slight issue that drop downs are too long and force the last option onto a new line. Try clinical findings as first option - the next box has very long option - Lymphadenopathy details - forces a new line for final box. Not sure how common this is though. Most other options are ok

PR #281, opened 2015-08-20
V0.4.2 318 extract search

fredkingham:
fixed the spacing so that we don't overlap.
we now use bootstrap tooltips and a new focus directive to drive the user through the flow.
saved searches are encapsulated on the right.
fixed the bug @GabPoll found

What's not fixed is some of the saved search naming stuff which should probably be part of a separate ticket.


davidmiller:
Looks great - much better than it was last week !

PR #288, opened 2015-08-26
684 csv new lines

fredkingham:
changes so that we use the native csv writer which by default outputs in the excel style of csvs.

Potentially more changes than were actually required... I've optimised a db query and changed to use dict comprehensions. We do seem to get a nice speed boost which is cool.

Also transferred the existing tests across and added some more. I think it potentially we could add even more, but this seems a good base.


GabPoll:
great work. is there a real world test for this you'd like us (@michaeledwardmarks) to try?

davidmiller:
@GabPoll I believe it's currently deplolyed to elcid04-test.

FYI I'm going to deploy this to ELCIDT today so that we can do some extracts for people :)

GabPoll:
Ah thank you, because we owe an extract on ID liaison. Also will serve as the best real world testing, better than heroku.

PR #300, opened 2015-09-23
701 patient detail view

fredkingham:


opal/static/js/opal/controllers/edit_item.js
```diff
@@ -78,18 +78,6 @@ angular.module('opal.controllers').controller(
 
 	    $scope.episode_category_list = ['Inpatient', 'Outpatient', 'Review'];
 
-        $scope.delete = function(result){
```
davidmiller: @fredkingham Why did we delete this ? Isn't this the place we want this? (related - the delete button doesn't seem to work currently on elcid04-test)

PR #313, opened 2015-10-04
312 svg for timeline

fredkingham:
we get a much cleaner timeline line with a background image


davidmiller:
@fredkingham is this deployed anywhere?

fredkingham:
so its deployed on the heroku instance at the moment, unfortunately although its better than what we've got at the moment, I'm not sure if its worth merging as its breaking on FF (not on Chrome) on https://elcid04-test.herokuapp.com/#/patient/78574392579438 (around the last <hr />)

fredkingham:
ok @davidmiller I think this works for all cases (not actually all, but all cases we expect to have)

PR #331, opened 2015-11-05
Invalidate on blur in the editing modal as well as the add patient

davidmiller:


opal/templates/modals/demographics_modal.html
```diff
@@ -2,49 +2,49 @@
...
-               </span>
-
-      </div>
+  </div>
+  <div class="form-group">
+    <label class="control-label col-sm-3">Date of birth</label>
+    <div class="col-sm-8">
+	  <input class="form-control" type="text"
+             name="date_of_birth"
+             ng-model-options="{ updateOn: 'blur' }"
```
davidmiller: Mostly whitespace, this is the interesting line...

PR #346, opened 2015-11-12
remove loading screen in print mode

fredkingham:


opal/static/css/opal.css
```diff
@@ -910,3 +910,9 @@ svg.speech-bubble-timeline-container{
     padding-left: 5px;
     text-align: left;
 }
+
+@media print {
+    .no-print{
+        display: none;
+    }
+}
```
davidmiller: I guess you've not seen this: 

https://github.com/openhealthcare/opal/blob/release-staging-10-09-2015/opal/static/css/print.css#L14

fredkingham:
good call, updated as suggested

PR #358, opened 2015-11-25
allow synonyms to be saved by the many to many and unit test accordingly

fredkingham:


opal/models.py
```diff
@@ -100,29 +100,50 @@ def get_field_type_for_consistency_token(cls):
...
-        if to_add_names:
-            field = next(i for i in self._meta.get_fields() if i.name == name)
-            to_add = field.related_model.objects.filter(name__in=to_add_names)
-
-            if not len(to_add_names) == len(to_add):
-                found_names = set(to_add.values_list("name", flat=True))
-                unexpected = list(set(to_add_names) - found_names)
-                error_msg = 'Unexpected fieldname(s): %s' % unexpected
-                logging.error(error_msg)
+    def get_lookup_list_for_names(self, lookuplist, names):
```
davidmiller: This get_lookuplist... function doesn't seem to return a lookuplist?

setup.py
```diff
@@ -38,7 +38,7 @@
         'requests',
         'django==1.8.3',
         'django-reversion==1.8.7',
-        'django-axes==1.3.9',
+        'django-axes==1.4.0',
```
davidmiller: This is something we'd already done elsewhere but just not set in setup.py ?

PR #366, opened 2015-11-30
bunch queries for content types

fredkingham:
local speed up

old
super GET /api/v0.1/options/ 200 (1.99 seconds) (365 SQL queries, 1869.0 ms)

new
super GET /api/v0.1/options/ 200 (1.13 seconds) (214 SQL queries, 1489.0 ms)

PR #367, opened 2015-11-30
877 schema loader parrelelise

fredkingham:


davidmiller:
@fredkingham This seems to have https://github.com/openhealthcare/opal/pull/366 in it as well - the promise stuff in the episode loader seems just fine - although Travis shows one of the javascript service tests failing :(

PR #368, opened 2015-11-30
optimise many to many fetches

fredkingham:


davidmiller:
Looks perfectly sensible to me - we never did this before because M2M is new.

Maybe this is a good time to add some test coverage of the serialisation ? - I couldn't find any :(

fredkingham:
I think some tests are done on a more granular level but I've added some higher scoped tests to make sure they're covered

PR #385, opened 2015-12-18
901 make lists declarative

fredkingham:


davidmiller:
Looking good ! 

Few notes as discussed earlier:

* Currently no tests in OPAL for `opal.core.patient_lists` 
* Currently no doc change to opal/doc/docs/guides/list_views.md 
* The flow through get, get_class isn't particularly obvious to me - you totally be right though ;) docs could make this just go away !
* The commented out stuff in Schema can go - honestly that should probably just become the backend for the /records/ API endpoint
* That means probably moving `opal.core.schemas.extract_schema` into `opal.core.search`
* Why is TaggedPatientList a mixin ? I think it can be a subclass and we let _itersubclasses follow the chain (I _think_ this just works - it does for `issubclass`)

At "some stage"

* Nothing in *.views.py should ever really return JSON as far as possible - everything in a *.api.py !
* We may well want to have a default application level schema in a setting.
* We may well want to separate out spreadsheety-lists from listy-lists

PR #386, opened 2015-12-18
Create a generic app loader

fredkingham:
to be merged in after 901 pull request


davidmiller:
@fredkingham LGTM - been thinking for a while that a generic version of this would be a useful thing to have.

- [ ] Suspect we might want some explicit tests for it though?
- [ ] I reckon a brief overview and some example code would be a useful thing to have - maybe here: http://opal.openhealthcare.org.uk/docs/guides/plugins/ ?

PR #393, opened 2016-01-11
New wardrounds

fredkingham:
note this depends on #386


fredkingham:
wrong version to synch with

PR #406, opened 2016-02-04
Series of updates to make the startproject script usable

davidmiller:
* Update detail templates so that we're overridable
* Add random data to opal
* don't cache templates in debug for generated apps
* have a base inpatient template that uses generated models


davidmiller:
BTW all of the landscape complaints are in create_random_data - and mostly about unused argument *args .... 

We should probably find a way to configure it to not care about that

PR #495, opened 2016-03-21
Bulk create

fredkingham:
create a bulk create api that can be used by upstream gloss and by pathways


opal/models.py
```diff
@@ -391,6 +393,49 @@ def get_active_episode(self):
...
+                          ],
+                      "investigation":[
+                            {"name": "some test", "details": "some details"}
+                          ]
+                     }
+        """
+        if "demographics" not in dict_of_list_of_upgrades:
+            if not self.id:
+                raise ValueError(
+                    "demographics are required when creating a new patient"
```
davidmiller: Is this the error message we actually mean here? Surely you're often not creating a new patient...

davidmiller: ...also, it's a method on an instantiated Patient - are you sure you want to always raise a valueerror if there's no demographics?

doc/docs/reference/patient.md
```diff
@@ -11,3 +11,9 @@ Returns a new `Episode` for this patient.
 Return the dictionary representation of this patient - suitable for serialization.
 
     patient.to_dict(user)
+
+
+#### bulk_update
+pass in a dictionary of subrecords you want to update, pass in an episode if one exists.
```
davidmiller: So... I would struggle to call this method based on this! A code snippet would be great, and mentioning that User is a required argument...

davidmiller: Probably worth adding Subrecord.bulk_update_from_dicts(parent, dicts, user) to subrecords.md as well?

opal/models.py
```diff
@@ -722,12 +769,51 @@ def get_display_template(cls, team=None, subteam=None):
...
+
+        for a_dict in list_of_dicts:
+            if "id" in a_dict or cls._is_singleton:
+                if cls._is_singleton:
+                    query = cls.objects.filter(**{schema_name: parent})
+
+                    if query.exists():
+                        subrecord = query.get()
+                    else:
+                        subrecord = cls(**{schema_name: parent})
```
davidmiller: Honestly, this shouldn't be possible... singletons should always be created at parent creation.... There's a case for this raising an error?

opal/models.py
```diff
@@ -722,12 +769,51 @@ def get_display_template(cls, team=None, subteam=None):
...
+                else:
+                    subrecord = cls.objects.get(id=a_dict["id"])
+            else:
```
davidmiller: This probably needs a test ? 
Maybe I'm missing it, but I can't seem to see anything that updates a previously existing non singleton subrecord?

PR #510, opened 2016-03-28
Infer label, model, lookuplist from a subrecord field

fredkingham:


davidmiller:
Chatting about this we reckon we should make the calling API be Django Python Class.field as a string rather than the front end version because this is not the front end.

davidmiller:
./bin/opal generate templates should use this syntax now

PR #511, opened 2016-03-29
add gloss allergy changes

fredkingham:
improve the url getting of record panels, add a lazy object to include all subrecords in templates


davidmiller:
- [ ] Change to be called "models"
- [ ] Delete some code ! (models is passed into some template contexts)
- [ ] Add the existence of this to some docs somewhere
- [ ] ./bin/opal generator script should add this context processor to default settings for a new app

PR #563, opened 2016-05-07
Add default diagnosis form for scaffolded applications

davidmiller:


fredkingham:
we only have an abstract diagnosis in OPAL, so we don't need this right?

PR #578, opened 2016-05-13
1051 investigations modal broken

fredkingham:
nuke the defaults after the test name is changed


fredkingham:
fixes openhealthcare/elcid#1051

PR #587, opened 2016-05-16
urls should work the same for modals as they do for records

fredkingham:


opal/urls.py
```diff
@@ -67,24 +65,10 @@
...
-# Generated subrecord template views
-for subrecord_model in subrecords():
-    sub_url = camelcase_to_underscore(subrecord_model.__name__)
-    urlpatterns += patterns(
-        '',
-        url(r'^templates/forms/%s.html/?$' % sub_url,
-            views.FormTemplateView.as_view(), {'model': subrecord_model},
-            name="form_template_view"
-            ),
-        url(r'^templates/modals/%s.html/?$' % sub_url,
```
davidmiller: What's happened to /templates/forms/foo ? Do we not use/need it? If so, do we still need the view ?

fredkingham:
good spot, also found another bug in the sourced from footer

PR #614, opened 2016-05-23
we need to make sure we use the tag that the episode has actually bee…

fredkingham:
…n tagged with in the growl alert


fredkingham:
I'm not wondering what happens when an episode has multiple tags....

PR #633, opened 2016-06-01
remove update on blur

fredkingham:


opal/templates/modal_base.html
```diff
@@ -14,7 +14,7 @@
 </div>
 <div class="modal-body" {% block modal_body_extra %}{% endblock %}>
   {% block modal_body %}
-    <form class="form-horizontal">
+    <form name="form" class="form-horizontal" novalidate>
       {% include column.get_form_template %}
```
davidmiller: Are we sure that this doesn't have unintended consequences in other places?

PR #682, opened 2016-06-12
allow for null booolean and datetime fields

fredkingham:


davidmiller:
+10 moving as much of the ./bin/opal stuff into opal.core as possible...

opal/core/scaffold.py
```diff
@@ -0,0 +1,71 @@
...
+    Given a RECORD, return it's relative template dir
+    """
+    modelsfile = inspect.getfile(record)
+    if modelsfile.endswith('.pyc'):
+        modelsfile = modelsfile.replace('.pyc', '.py')
+    appdir = ffs.Path(modelsfile)[:-1]
+    templates = appdir/'templates'
+    return templates
+
+def create_template(record, scaffold_base, template_type):
```
davidmiller: Is this function used anywhere?

fredkingham:
I think its used by create_display_template_for and create_form_template_for

I think there maybe some cleaning up we could do to remove duplication between the functions but best to get them all in at least a testable place

davidmiller:
Sorry - meant the `def create_template(record, scaffold_base, template_type):` one below - it's just a pass?

fredkingham:
are sorry, that feels like something I started doing as a refactor, then
figured that it'd be best to just get everything copied across and unit
tested before any further steps.

(Not sure a refactor is a priority, its all pretty good)

PR #692, opened 2016-06-18
the key work model is already take, use a different one

fredkingham:


fredkingham:
using a template tag instead

PR #701, opened 2016-06-21
refactor our datetime and date setting for use in super

fredkingham:


fredkingham:
killing this in favour of fix-date-time-for-patient-consultation2

PR #702, opened 2016-06-21
Experimental alternative datey thing

davidmiller:
This is what I would probably naively do here... it's mostly splitting the deserialization of date[time] strings apart from the separate action of setting that on the appropriate model... entirely possible that what I'm missing is one of the "But because Many To Many/complex models" things that I haven't gotten my head around yet - but I still haven't :)

### DIFF
--- a/opal/models.py
+++ b/opal/models.py
@@ -35,20 +35,16 @@ def get_default_episode_type():
     return app.default_episode_category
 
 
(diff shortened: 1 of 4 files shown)


fredkingham:
killing this in favour of a pull request straight onto 0.7.0

PR #721, opened 2016-06-24
added a query set method that does the basic searching

fredkingham:


davidmiller:
This is great !
(And not just the French ;) )

Could we have something about the custom manager and the search method in `./doc/docs/reference/patient.md` and `CHANGELOG.md` ?

fredkingham:
doh! Not only does it need more tests, it needs tests fixed, gotta run will try and get this done tonight

PR #755, opened 2016-07-05
750 validation for dates number

fredkingham:


opal/static/js/opal/services/fields_translater.js
```diff
@@ -62,17 +62,31 @@ angular.module('opal.services').service('FieldTranslater', function($rootScope){
     return subrecord;
   }
 
+  this.cleanEmptyStrings = function(fieldValue){
```
davidmiller: I expect a thing called `cleanEmptyStrings` to conceptually be able to handle multiple strings - by clean you mean trim?

opal/static/js/opal/services/fields_translater.js
```diff
@@ -62,17 +62,31 @@ angular.module('opal.services').service('FieldTranslater', function($rootScope){
...
             }
-            return fieldValue;
         }
-        else if(fieldMapping.type == 'date_time'){
-            return moment(fieldValue).format(DATETIME_FORMAT);
+        else if(fieldMapping.type === 'date_time'){
+          fieldValue = moment(fieldValue).format(DATETIME_FORMAT);
+        }
+        else if(fieldMapping.type == 'integer' || fieldMapping.type == 'float'){
+          fieldValue = self.cleanEmptyStrings(fieldValue);
```
davidmiller: Why just integers? 

Date & datetime a subsequent PR ?

fredkingham:
so we're now checking datetimes and dates. Datetimes previously were always assumed to be moments so we're now not doing this in case we have any special cases like date of birth was for date.

I changed it from clean empty strings to clean strings, as we can have multiple empty bits, but I use clean because if there's no character that isn't it space it returns and undefined. Not sure, naming things is hard.

PR #767, opened 2016-07-08
we definitely want to not have black text in the header

fredkingham:
do we want links in panels to go orange? I'm not sure, I think we want something and I think this is better than an underline

### DIFF
--- a/opal/static/css/opal.css
+++ b/opal/static/css/opal.css
@@ -129,6 +129,15 @@ h1 small { color: #eff; }
 }
 .panel-container { margin-top: 20px; }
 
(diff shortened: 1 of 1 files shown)


GabPoll:
which one is the header? - screenshot?

fredkingham:
the top of the pathway is become a link back to the patient, at the moment all links wherever they are are black, so this includes a change to make links white if they're on the green background.

The next question is how do we signal to the user that they're a link, I've changed it so they go orange on mouse over, I'll push to somewhere later and you can try it... My ability to make things look nice is terrible

screen shot 2016-07-08 at 16 33 28

PR #774, opened 2016-07-08
750 backend validation

fredkingham:


fredkingham:
@davidmiller I think this might be good to go

PR #809, opened 2016-07-30
Sortable features shouldn't allow multiple of the same value of order…

### LINKED ISSUE #653: Add SortableFeature Validation
We shouldn't really allow multiple features to declare the same order number - let's prevent this with a validator.

davidmiller:
…. Make that fail is_valid

closes #653

### DIFF
--- a/opal/core/discoverable.py
+++ b/opal/core/discoverable.py
@@ -1,6 +1,9 @@
 """
 OPAL utilities for discoverable functionality
 """
+import collections
+import inspect
+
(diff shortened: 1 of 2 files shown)


fredkingham:
closing this because I think we have an application that depends on 2 different plugins that implement sortable, in my application I don't want them to blow up.

This may be an argument against the person who wrote the plugin, or it may be an absurd edge case, but we'll close for the moment until not having this feature causes us an issue.

PR #819, opened 2016-08-10
a clip board directive that is similar to ngClipboard, but will use g…

fredkingham:
…rowl by default to give user feedback


davidmiller:
@fredkingham looks good - could we have something in http://opal.openhealthcare.org.uk/docs/reference/javascript_helpers/#angular-directives with a usage example and something in the changelog ?

fredkingham:
good call

fredkingham:
documentation added

PR #827, opened 2016-08-14
This raise is unreachable?

davidmiller:
I can't think of a way this raise is actually a reachable code path ?
Am I mis-reading ? 
(If so, please tell me how ! :D )

If not can we just remove this line?
I've walked right back through the history of the file and it seems to have always been there...

### DIFF
--- a/opal/utils/__init__.py
+++ b/opal/utils/__init__.py
@@ -36,7 +36,6 @@ def stringport(module):
                 raise ImportError(msg)
         except ImportError, e:
             raise ImportError(msg)
-        raise ImportError(msg)
 
 
 def _itersubclasses(cls, _seen=None):


fredkingham:
Doesn't it change the error message thrown by the import lib 5 lines above? (Not sure what the message would be otherwise or if this is helping)

davidmiller:
@fredkingham I thought it was changing say, errors thrown by L31-34 into `ImportError`, but then I did this:

```
>>> from opal.utils import stringport
>>> class SillyString(str):
...   def rsplit(self, splitter, maxlen):
...     raise Exception('Get in the sea.')
...
>>> stringport(SillyString('foo.bar'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "opal/utils/__init__.py", line 31, in stringport
    module, obj = module.rsplit('.', 1)
  File "<stdin>", line 3, in rsplit
Exception: Get in the sea.
```

Still not entirely sure I'm not missing something though...

PR #833, opened 2016-08-18
Remove options

davidmiller:
This finally removes the Options API/Service entirely.


fredkingham:
we also probably want to kill
opal/opal/scaffolding/plugin_scaffold/runtests.py.jinja2:                   OPAL_OPTIONS_MODULE = '{{ name }}.tests.dummy_options_module',

then we need to remember to remove this from every projects run tests right? (Not suggesting we do it right now)

davidmiller:
Good shout about the plugin test runner - that should be gone now...

PR #835, opened 2016-08-19
all template and a json requests should be strictly enforcing a logge…

fredkingham:
…d in user


opal/core/api.py
```diff
@@ -371,16 +373,3 @@ def post(self, *args, **kwargs):
             episode.set_tag_names(current_tags, None)
         resp = {'ok': 'Got your referral just fine - thanks!'}
         return _build_json_response(resp)
-
-
-class EpisodeListApi(View):
-    """
```
davidmiller: @fredkingham Do you think this is not used anywhere (I can completely believe that it's not, just haven't checked myself...)

davidmiller:
In general this looks great - I reckon it needs entries in  

* Changelog
* Upgrading (Suggest that applications or plugins defining APIs swap out Viewset for Loginrequired)
* http://opal.openhealthcare.org.uk/docs/guides/plugins/
* http://opal.openhealthcare.org.uk/docs/guides/json_api/

opal/core/api.py
davidmiller: It's not even wired up to a URL so I assume we're fine.

fredkingham:
documentation added

PR #838, opened 2016-08-24
we add the index to the id, if its not in a for loop this works fine,…

fredkingham:
… if it is in a for loop this works fine, we maybe want to let a model id argument be passed in but we have no use case for this at the moment.

Changed model to modelname as model has "." in it that are unexpected from an id field


fredkingham:
fixes openhealthcare/uclp-tb#208

PR #847, opened 2016-08-30
docs(README.md) correct naming of AngularJS

d2s:
- Remove unneeded whitespace from name of a software library.


davidmiller:
Thanks @d2s - that's how they spell it, so I guess we should too !

PR #848, opened 2016-08-30
Use ng idle

fredkingham:


davidmiller:
@fredkingham is this still something we want/need? lots of conflicts now...

davidmiller:
After discussing: We're going to come back to this branch after completing our UCH investigation.

PR #852, opened 2016-09-01
Remove static

fredkingham:


davidmiller:
@fredkingham is inpatient_admission_form.html used anywhere?

fredkingham:
so no, there's a record panel in elcid but its editable = False.

But its a model in OPAL and our tendency is therefore to provide a form...

PR #859, opened 2016-09-03
856 nothing but staic

fredkingham:


davidmiller:
Looks good to go as long as we can pull the static argument out of the reference docs !

fredkingham:
The day is coming where I remember to do that!

PR #862, opened 2016-09-11
Load reference data from individual apps

fredkingham:


davidmiller:
Looks good, 

couple comments: 

- Plugin.directory() should probably go into our docs
- The tutorial is still using the old load_lookuplists -f command
- The Command.handle() method should probably look in OPAL and the current Application as well as plugins

fredkingham:
depends on https://github.com/openhealthcare/opal/pull/864

PR #865, opened 2016-10-10
form name is now form by default. This means that things just work in…

fredkingham:
… modals and pathways. You can override it if you want, but the only reason you'd want to is if we had multiple forms on the same page, which is not usecase we have at the moment

### DIFF
--- a/opal/templates/modal_base.html
+++ b/opal/templates/modal_base.html
@@ -14,7 +14,7 @@
 </div>
 <div class="modal-body" {% block modal_body_extra %}{% endblock %}>
   {% block modal_body %}
-    <form name="form" class="form-horizontal" novalidate>
+    <form name="form" ng-submit="" class="form-horizontal" novalidate>
       {% include column.get_form_template %}
     </form>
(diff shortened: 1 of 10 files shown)


opal/templates/modal_base.html
```diff
@@ -34,8 +34,7 @@
   {% endif %}
 
   {% block modal_save %}
-  <button one-click-only class="btn btn-primary" ng-click="save('save')"
-          ng-disabled="form.$invalid || saving">
+  <button check-form="form" class="btn btn-primary" ng-click="form.$valid && save('save')">
     <i class="fa fa-save"></i>
```
davidmiller: @fredkingham Does this button no longer need the `one-click-only` directive?

fredkingham: I've rolled the one-click only logic into the check-form directive as we never want to check a form without the one click.

fredkingham:
I'm actually going to play with this more, it works fine in modals, just want to make sure it works with all the pathways before we merge

PR #883, opened 2016-11-08
Remove Angular Strap

davidmiller:
This is the entry PR for Removing Angular Strap.
It updates all OPAL usage of the following UI widgets: 

* bs-tooltip
* bs-popover
* bs-typeahead
* bs-datepicker
* bs-timepicker

It also removes old versions of libraries, and updates UI-Bootstrap to a version for which the documentation is at least hosted in their archive (http://angular-ui.github.io/bootstrap/versioned-docs/0.14.3/)

We are pausing at 0.14.3 because after this our $modal usage will raise an error and we will need to update all places where we call `$modal.open` or `$modalInstance.x`. This PR seems large enough as it is.

An open question is whether to update all $modal use across all OPAL apps/plugins, or to alias it for now.

There are also pending PRs in the following repo/branches to be opened once this is merged: 

* elcid/860-remove-angular-strap
* uclp-tb/860-remove-angular-strap
* elcid-uch/860-remove-angular-strap
* obs/860-remove-angular-strap


fredkingham:
we've removed min date, which is used in opal/opal/templates/_includes/discharge_form.html, it seems sensible but if angular strap doesn't give us the feature we should remove it in both places

fredkingham:
apart from the above, looks great, lets heroku it up for live tests

davidmiller:
UCH: http://elciduch-strapless.herokuapp.com/#/list/infectious_diseases-id_inpatients
RFH: http://elcidrfh-strapless.herokuapp.com/#/
TB: http://tb-strapless.herokuapp.com/#/

fredkingham:
we've sometimes lost the animation on the modal (not sure if this is a bug). Not always, but always if you refresh the page, first time you click on add patient (weird).

I think we should restyle the date widget, at least to use elcid colours.

tb date of referral is using the old date picker, and so does some other spots, but I think we knew that

The time picker in clinical advice and possibly other places needs tweaking

screen shot 2016-11-14 at 11 26 33

Apart from that looks good (I found lots of other unconnected issues unfortunately...)

davidmiller:
TB Test server seems to be pointing at the wrong OPAL branch...

davidmiller:
I've updated the datepicker to hide things and look a bit nicer...
The TB test server was using the wrong OPAL branch, but can't currently deploy because lab test errors as part of the app compilation...

PR #894, opened 2016-11-15
don't enforce validation, if they don't want to have a form, don't ma…

fredkingham:
…ke them, we're all friends here


davidmiller:
@fredkingham I think we decided to move to an approach based on splitting out the templates and forcing users to manually e.g. explicitly remove validataion ?

fredkingham:
yep sorry this pull request has been superceded and is dead

PR #899, opened 2016-11-19
322 admin urls

davidmiller:


fredkingham:
looks great, we need to update tb that overrides the nav bar

PR #902, opened 2016-11-20
887 remove team model

davidmiller:
Removes the `opal.models.Team` thing that is neither used nor useful.

Once merged, this will trigger two subsequent PRs which have branches waiting: 

* elcid/887-remove-team-model
* elcid-rfh/887-remove-team-model


davidmiller:
This is that same thing of deleting tested code and then coveralls complaining...

Let's get the other three merged and then see where we are with test coverage...

fredkingham:
yep, we need to remove from the admin.py, not urgent as its commented out,

opal-wardround uses it in at least
opal-wardround/wardround/tests/test_team_filter.py

opal-referral uses it in at least
opal-referral/referral/tests/test_api.py:from opal.models import Patient, Episode, Team
elcid/research/teams and tests
elcid/walkin/views.py

davidmiller:
@fredkingham do you want to look at https://github.com/openhealthcare/elcid/compare/887-remove-team-model and https://github.com/openhealthcare/elcid-rfh/compare/887-remove-team-model 

I believe these deal with elCIDs, and are ready to go once it's gone from Opal.
(They remove both references to teams and change add_episode_modal urls).
I can split out the team removal stuff, but we'll still need another PR post merging this one anyway, so not sure if that's really useful...

fredkingham:
Removing a Team will remove a Tagging row that is connected to it. 

Can we check that if we have a Tagging row that is still attached to a Team and we drop the column that this doesn't drop related Tagging rows. It shouldn't I don't think, but lets double check it.

Also we may need to update the updating document with information about this.

PR #909, opened 2016-11-29
config should be stored just under the route dir

fredkingham:


davidmiller:
@fredkingham Random test failures !

Pun intended.

PR #914, opened 2016-12-26
Tabbed list groups

davidmiller:
This PR gives us the long-requested ability to group sets of PatientLists for a single clinical service as tabs across the top of a spreadsheet list, giving  faster access than the list switcher dropdown.

To see it in action, checkout the `tabbed-list-groups` opal branch, as well as the `tabbed-list-groups` elCID@UCH branch, and view the OPAT/Walkin lists.


opal/views.py
```diff
@@ -71,6 +71,14 @@ def get_context_data(self, **kwargs):
         context['list_slug'] = list_slug
         context['patient_list'] = self.patient_list
         context['lists'] = PatientList.for_user(self.request.user)
+
+        context['list_group'] = None
+        if self.patient_list:
```
fredkingham: Will there ever not be a patient list at this point?

opal/core/patient_lists.py
```diff
@@ -119,6 +119,69 @@ def get_template_prefixes(self):
...
+    @classmethod
+    def get_member_lists(klass):
+        """
+        A hook for dynamically customising the members of this list group.
+
+        Returns an iterable of PatientLists
+        Defaults to the `.member_lists` property
+        """
+        for l in klass.member_lists:
+            yield l
```
fredkingham: should we add user as an optional argument here? Either that or just change klass.member_lists to be a tuple and then just return klass.member_lists

opal/core/patient_lists.py
```diff
@@ -119,6 +119,69 @@ def get_template_prefixes(self):
...
+            if l.visible_to(user):
+                yield l
+
+    @classmethod
+    def visible_to(klass, user):
+        """
+        Predicate function to determine whether this list is meaningfully
+        visible to this USER
+        """
+        if len(list(klass.get_member_lists_for_user(user))) > 1:
```
fredkingham: we don't need len()

opal/templates/patient_lists/tabbed_list_group_base.html
```diff
@@ -0,0 +1,11 @@
+<div class="patient-list-group">
+  <ul class="nav nav-tabs nav-justified" role="tablist">
+    {% for member in members %}
+    <li role="presentation" {% if member.display_name == active_list.display_name %}class="active"{% endif %}>
+      <a href="#/list/{{ member.get_slug }}" aria-controls="home" role="tab" data-toggle="tab">
```
fredkingham: is this actually aria-controls="home" (genuinely no idea, even after a quick google and have to head out in a mo sorry!)

opal/views.py
```diff
@@ -71,6 +71,14 @@ def get_context_data(self, **kwargs):
...
+            group = TabbedPatientListGroup.for_list(self.patient_list)
+            if group:
```
fredkingham: this is pretty much just duplicating get_member_lists_for_user right?

fredkingham:
so I think this is great, I could merge as is. I'm not sure why you're not passing user as an optional argument to the get_lists, that way you can remove some duplication.

I've put some nit picky comments, feel free to ignore and merge if you think they're too nit picky. They aren't show stoppers

opal/core/patient_lists.py
davidmiller: Don't need in terms of "there's an alternative syntax" ? 

You could spell it like this I guess....
```python 
return bool(list(klass.get_member_lists_for_user(user)))
```

... my reckons would be that the conceptual criteria we're using for visibility is explicitly the length of the thing, so doing it explicitly gives you better readability.


TBH the thing I'd question about this would be if 1 is a less sensible default than 2 here - e.g. is a tabbed list group of two things useful enough to clutter the ui

opal/templates/patient_lists/tabbed_list_group_base.html
davidmiller: I have literally no idea - I cargo culted it from the example markup here here: http://getbootstrap.com/javascript/#tabs

opal/views.py
davidmiller: If you hit say http://localhost:8000/templates/patient_list.html/opat-opat_referrals_or_something_made_up 

We check for the existence of `self.patient_list` in a bunch of places in the View - there's a good case for a PR that changes `dispatch` to raise a 404 instead of explicitly setting `None` but that's a whole different PR :)

opal/views.py
davidmiller: Not entirely sure what you mean ? 

 `member_lists_for_user` gets the iterable of `PatientLists` that make up this group, `group.visible_to` determines whether the group itself is visible to the user in question, `for_list` figures out if this List is in a group.

The difference is useful because for instance I might want to restrict who can see the group UI - for instance I have three different team lists, and all doctors can, if the choose, navigate to all three lists, but will be mostly working from one. However the admin/nursing staff cover all three teams because they're on the same physical ward and so move in between all day. I can do this by overriding `visible_to` to take into account user roles etc

opal/core/patient_lists.py
davidmiller: > should we add user as an optional argument here?

And collapse the two methods `get_member_lists` and `get_member_lists_for_user` into one presumably? 
You could obv... I'm not sure what it would get you TBH ?

I guess my argument would be that it gives you slightly clearer hooks to override in your implementation - this method just gets members, and doesn't care about permissions / visibility / who, this other method cares about determining which of the members can be seen by this person.

Not sure it makes a huge amount of difference either way...

> change klass.member_lists to be a tuple 

Not exactly sure what you're getting at here? 

The original implementation of this was to just return `klass.member_lists`... 
I changed it to be a generator for some sense of internal API consistency once the class had grown `get_member_lists_for_user` which iterated over things and returns a generator.

In general I don't have particularly strong feelings about returning lists vs. generators, I tend towards implementing generators and letting calling code cast to a list if it's going to iterate through multiple times for some reason.

davidmiller:
Thorough reviews are good !

(Although I think there's less duplication than you reckon ;D )

I'm in no mad rush so I'll wait for your responses !

opal/views.py
fredkingham: sorry, my mistake

PR #944, opened 2017-01-23
if there is no episode in scope, make sure we add it to the scope

fredkingham:


davidmiller:
This is about altering the scope of `episodePresent` ?
How does this manifest itself in applications ?

fredkingham:
the move of episodePresent is cosmetic, I'm just hoisting the var declaration to the top of the function scope

PR #953, opened 2017-01-28
more may be required, but this is the minimum

fredkingham:


davidmiller:
I don't think the original version is actually what happens?

PR #971, opened 2017-02-06
Users for menuitems

davidmiller:
To allow filtering of Application menu items by user we need to pass the user into the `get_menu_items` method.

Also does some minor formatting and adds a Flake8 config that ignores some paths and Error classes.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,4 @@
+# -*- mode: conf -*-
+[flake8]
+ignore  = E203,E221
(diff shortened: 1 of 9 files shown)


fredkingham:
looks good, user is in the context, so changed to use it from there instead, I'm happy to merge if @davidmiller you're happy

 https://docs.djangoproject.com/en/1.10/ref/templates/api/#django-contrib-auth-context-processors-auth

PR #973, opened 2017-02-07
First implementation of semi-working Discoverable Managers

davidmiller:
Note that the target branch for this PR is the feature branch not the release branch

This contains the first implementation of a Discoverable Manager.

Code review for this PR is intended to explore potential problems with this implementation - situations and use cases not tested, potential complications et cetera.

Recommendations for extra tests particularly 

#### Open Questions

How to do the following:

* Discoverable Validation
* Restrictable Features
* Sortable Features

### DIFF
--- a/opal/core/discoverable.py
+++ b/opal/core/discoverable.py
@@ -5,7 +5,9 @@ from django.conf import settings
 from six import with_metaclass
 
 from opal.core import exceptions
-from opal.utils import camelcase_to_underscore, _itersubclasses, stringport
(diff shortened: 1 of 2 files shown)


fredkingham:
I think this looks great, I'm not sure that we want all AbstractBases to definitely be discoverables, I *think* what we actually want is for a discoverable class to be able to exist in the children of another discoverable class.

Obviously this could case the world to blow up.

But if we can stop the world blowing up. This is potentially what we want

PR #993, opened 2017-02-27
982 change extract form field appearance

fredkingham:
lots of space changes, this is what we're looking like now

screen shot 2017-02-27 at 19 09 19


fredkingham:
note I've left in the js for the saved searches because I don't think we're definitely removing this feature right?

davidmiller:
This looks great :)

Most of the extract.js changes are just whitespace/indentation but some refactoring makes it look like everything has changed ?

PR #994, opened 2017-02-28
There is no need to register Plugin ,It is a discoverable features now.

khchine5:


fredkingham:
:thumbsup: :thumbsup: :thumbsup:

Great call!

PR #995, opened 2017-02-28
Remove unused import

khchine5:


davidmiller:
Thanks @khchine5 !

Good spot :)

I was thinking the other day that we don't test the generated code particularly well in the build. 
There are probably several things in the generated code that would generate e.g. Flake8 errors!

PR #996, opened 2017-03-02
patient service must only get the record editor from initialised epis…

fredkingham:
…odes


fredkingham:
Merging as this is required for testing a release

PR #998, opened 2017-03-03
fix the casting of an episode date of admission when you add an episode

fredkingham:


davidmiller:
In the background, I'm a little concerned that we have `var DATE_FORMAT` hard coded in three places again.
(Although that's definitely not something to include in #998 !)
It might be worth pulling some settings from settings into a metadata thing.
This might be another case: https://github.com/openhealthcare/opal/blob/7c7eaf03f632c6112ac5eb6295e20349bd8ce862/opal/templates/opal.html#L34

PR #1011, opened 2017-03-08
Refactor opal.core.api to use json_response consistently. This ensure…

davidmiller:
…s that we will always use the OpalSerializer and that we don't have to decide which to use in opal.core.api - we can follow the convention of the code around us.


opal/core/views.py
```diff
@@ -49,6 +49,7 @@ def json_response(data, status_code=200):
     """
     response = HttpResponse()
     response['Content-Type'] = 'application/json'
+    response.data = data
```
fredkingham: what does this do?

PR #1016, opened 2017-03-09
Update episode to be prototypical

davidmiller:


fredkingham:
@davidmiller can we put this into 8.2?

PR #1021, opened 2017-03-09
choices should be ordered appropriately

fredkingham:


davidmiller:
This should maybe go into 0.8.2 ?

PR #1026, opened 2017-03-12
Get full name

davidmiller:
It's a little silly to construct the full name in multiple controllers...


fredkingham:
this is an improvement, not sure if we should have this on patient, but I guess that's a different question

PR #1027, opened 2017-03-12
Userprofiles0

davidmiller:
Sometimes it's nice to expose certain details about users of the system - it allows us to humanise things... 

This PR creates a `/user/ ` API endpoint, and provides some helpers for interacting with that service in the UI. 

That allows us to implement things like say, 'who last edited this record' with something like the following in `modal_form_base.html`:

```html
<div ng-show="item.created" class="content-offset-20">
  <p class="text-muted text-center">
    <span ng-hide="item.updated">
      <img class="text-avatar" avatar-for-user="[[ item.created_by_id ]]">
      Created by
      <span full-name-for-user="[[ item.created_by_id ]]"></span>
      [[ item.created | shortDate]]
    </span>
    <span ng-show="item.updated">
      <img class="text-avatar" avatar-for-user="[[ item.updated_by_id ]]">
      Last updated by
      <span full-name-for-user="[[ item.updated_by_id ]]"></span>
      [[ item.updated | shortDate]]
    </span>
  </p>
</div>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+### 0.8.2 (Minor Release)
+
+### User data for the client
+
+Adds a `User` Angular service that enables applications to use user data.
+Also adds some directives to make it easy to render User names and avatars.
+
 ### 0.8.1 (Minor Release)
 
 #### Cookies for the future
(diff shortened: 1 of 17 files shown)


fredkingham:
@davidmiller yes yes yes and 100x yes, but...

conflicts.

I think we should be showing a user icon until the promise has loaded? Isn't the text going to jump about a bit otherwise? 

We need some form of help text that lets the user know how to create gravatars.

My suggestion would be if you click on the gravatar it brings up a user modal maybe, with a description, email of the user and a how do I add a gravatar link at the bottom?

I think all of the above should be follow ups, happy to merge this after conflicts are resolved, but we should ticket them up. (Or not if you don't agree)

PR #1039, opened 2017-03-16
985 pull through defaults help records

fredkingham:
we are now using our search style sheet for search specific stylz

 change the front end logic to give us the ability to populate it


davidmiller:
@fredkingham looks great - have added some documentation for new filters and merged :)

PR #1045, opened 2017-03-19
Use <input type="number"> for numbers because they're numbers

fredkingham:
so in theory this is the "right" thing to do, in practice, it would stops numeric fields having any letters in them (although bizarrely not e). Its only implemented in Chrome as far as I can tell..

those are the cons. The pros are that if its on a mobile/ipad you get a numeric number plate come up instead of the keyboard which is quite cool.

I'm guessing this is a no go because people would find it confusing... but its a talking point


davidmiller:
LGTM if there were tests & docs...

fredkingham:
@davidmiller I think that's added

davidmiller:
@fredkingham I think you have merge conflict artifacts left in the CHANGELOG - don't suppose you could take another look at making sure it's merging the right things etc?

PR #1047, opened 2017-03-21
First pass adding some human readable help text for abstract model fi…

davidmiller:
…elds


fredkingham:
@davidmiller at the moment, I think the front end uses lower case true/false. I think we should probably capitalise the front end right? Just we need to be consistent with the help text

davidmiller:
It's lower case in the widget, and uniformly Capitalised in text... also, Coveralls is a liar. Merging.

PR #1049, opened 2017-03-21
970 menuitems

davidmiller:


fredkingham:
change it from classes to class (or make it a list)

also inline generators (i for i in [1, 2, 3]) are a bit cleaner than declaring a function

PR #1051, opened 2017-03-24
Tweak the display of lookuplists in extract 

davidmiller:
open a modal with the reference data in it.


fredkingham:
This is great!

Should we include the lookup lists in zip file?

PR #1052, opened 2017-03-24
987 generate documentation for fields

fredkingham:


opal/models.py
```diff
@@ -122,6 +122,52 @@ def _get_field_type(cls, name):
         )
 
     @classmethod
+    def get_human_readable_type(cls, field):
```
davidmiller: This is ++ useful, presumably at some point for the in-line documentation on search / extract pages?

I'm a little concerned about just how much we hoist onto our models though... an alternative implementation here would be to have this as a function that takes the actual field in e.g. `opal.models.schema` ?

As a minimum I think we'd want to document the method/function existing, and if it is going to be a classmethod that takes a string field name then we might rename the argument?

opal/models.py
```diff
@@ -122,6 +122,52 @@ def _get_field_type(cls, name):
...
+            models.IntegerField,
+            models.FloatField,
+            models.DecimalField,
+        )
+        if isinstance(field_type, numeric_fields):
+            return "Number"
+
+        if isinstance(field_type, ForeignKeyOrFreeText):
+            t = "Any result is possible, but a curated list of {} is suggested"
+            return t.format(field_type.foreign_model._meta.object_name)
```
davidmiller: I think this is out of sync with what the front end says in 0.8.2 ?
(In fact I think you've not merged this branch with0.8.2 since the lookuplist modal thing in https://github.com/openhealthcare/opal/pull/1051 )

opal/models.py
```diff
@@ -176,42 +222,44 @@ def get_field_enum(cls, name):
             return [i[1] for i in choices]
 
     @classmethod
+    def build_schema_for_field(cls, fieldname):
+        getter = getattr(cls, 'get_field_type_for_' + fieldname, None)
```
davidmiller: ++ good idea to split this out, although again there is a case for moving this functionality to `opal.core.schemas` if we're moving it around ?

I think we might like to have a referencedata documentation entry for it and at least some unittests for this individually ?

opal/core/search/extract.py
```diff
@@ -107,6 +149,11 @@ def zip_archive(episodes, description, user):
         make_file_path = functools.partial(os.path.join, target_dir, zipfolder)
         zip_relative_file_path = functools.partial(os.path.join, zipfolder)
 
+        file_name = "data_dict.html"
```
davidmiller: Could this become `data_dictionary.html`  ?
(I'm not sure dictionary -> dict is a common contraction outside of Python programmers :) )

opal/models.py
fredkingham: I mean yes and no :), so for lab tests and anything complicated we probably would want to manually override this.

What would you feel about moving the whole SerialisableFields model into a discoverable? That had defaulted to a generic Schema that takes the model as an argument. It would move all this logic out of the model while still allowing it to be overridable.

I think this is quite a good idea... However I don't want to do it now so I think leaving it where it is is more consistent.

Changing it to field_name and adding documentation sounds like a good idea

fredkingham:
@davidmiller what are we missing on this jobby?

PR #1056, opened 2017-04-04
Bring through symptoms patient

fredkingham:
This is a little bit much.. Essentially it was on the thinking that we want some sort of serialisation into different formats.

It gives us display names in the headers and we have the patient id in every row obviously. It also lets us pass in fields for the next stage.

- Display name as col headers
- Symptoms M2m in extract
- Use to_dict to re-use getters
- Pass in fields to render as a csv
- Adds patient_id to every row


fredkingham:
closing this, as its been broken down into smaller requests

PR #1057, opened 2017-04-05
episode-fields-to-extract (extract part 1)

fredkingham:
Hopefully a bit of a no-brainer. It think we want to look as whether we want this to be on the model, but for now, consistency will do


davidmiller:
This seems like the right thing to do - have added an explicit test for `opal.core.search.extract.episode_csv` which as far as I can tell isn't currently explicitly tested...

PR #1084, opened 2017-04-19
custom-query

fredkingham:
If we're happy with this, when its merged, I'll move the extract schema end point out of the opal schema and into opal search


opal/core/schemas.py
```diff
@@ -4,6 +4,7 @@
 import itertools
 from opal.core.subrecords import subrecords
 from opal import models
+from opal.core.search.search_query import SearchQuery
```
davidmiller: I think this should move now tbh

opal/core/search/queries.py
```diff
@@ -10,6 +10,7 @@
 from opal import models
 from opal.core import fields, subrecords
 from opal.utils import stringport
+from opal.core.search.search_query import SearchQuery
```
davidmiller: `opal.core.search.queries` and `opal.core.search_query` ? 

I guarantee I won't remember the distinction in 6m :)

davidmiller: I've been using the word rule...

opal/core/search/search_query.py
```diff
@@ -0,0 +1,114 @@
+from opal.core.discoverable import DiscoverableFeature
+from opal.utils import camelcase_to_underscore
+from opal.models import Episode, deserialize_date
+
+
+class SearchException(Exception):
+    pass
```
davidmiller: Can we inherit from `opal.core.exceptions.Error` if we need a custom exception?

opal/core/search/search_query.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class SearchQueryField(object):
+    lookup_list = None
```
davidmiller: I think I'd definitely like documentation for this :)

opal/core/search/search_query.py
```diff
@@ -0,0 +1,114 @@
...
+    def get_fields(self):
+        return self.fields
+
+    def to_dict(self):
+        return dict(
+            name=self.get_slug(),
+            display_name=self.display_name,
+            advanced_searchable=True,
+            fields=[i().to_dict() for i in self.get_fields()]
+        )
```
davidmiller: I don't really see why you need to instantiate these... particularly given it means they then work differently to the discoverable methods ?

davidmiller:
Merge conflicts :(

davidmiller:
@fredkingham More merge conflicts !

PR #1085, opened 2017-04-20
location detail should be looking at item not row right?

fredkingham:


davidmiller:
It's definitely doing this deliberately.

`date_of_admission` is not a field on `Location`.

This is prob not the correct way to access `Episode.date_of_admission` - for instance, will only work in list view, but changing it to `item.date_of_admission` breaks even that :)

I think there's a reference to the relevant episode on the Item which might be better...

fredkingham:
ah ok, makes sense, kill it

PR #1093, opened 2017-04-26
detect changes in radio buttons and text areas

fredkingham:


davidmiller:
This looks very sensible :)

Probably worth updating the forms templatetags documentation to explain the change arg?

Not sure, but I could def. be convinced that we could use a test to make sure that change flows through...

fredkingham:
Good call, documentation and tests should make sure this doesn't happen again

PR #1100, opened 2017-05-02
allow searching via integers/floats etc

fredkingham:


opal/core/search/templates/extract.html
```diff
@@ -4,7 +4,7 @@
       <div class="panel-heading">
         <h2>
           <i class="fa fa-search"></i>
-          Patient Search
+          Patient Search hello
```
davidmiller: I think this may be a debugging artifact? ;)

fredkingham:
How presumptive!

PR #1103, opened 2017-05-02
you need to be able to search by synonym

fredkingham:


opal/tests/test_core_search_queries.py
```diff
@@ -187,6 +187,28 @@ def test_episodes_for_date_fields_after(self):
     def test_episodes_for_m2m_fields(self):
         criteria = dict(
             column='hat_wearer', field='Hats',
+            combine='and', query='Derby', queryType='Equals'
```
davidmiller: Aren't these the wrong way round ? e.g. the test for synonym is called for_m2m and the test for m2m synonyms is m2m_fields_with_synonyms ?

opal/core/search/queries.py
```diff
@@ -178,27 +178,35 @@ def _episodes_for_date_fields(self, query, field, contains):
...
+            return
+
+    def _episodes_for_fkorft_fields(self, query, field, contains, Mod):
+        model = get_model_name_from_column_name(query['column'])
+
+        not_synonym = self._get_name_from_synonym(
+            getattr(Mod, field).foreign_model, query['query']
+        )
+
+        name = not_synonym or query['query']
```
davidmiller: You're _sure_ you want to do this rather than explicitly check to see if the return value is None ? :) 

I _think_ I naturally would have written that "more stupid" (TM) ;)
e.g.
```python
name = query['query']
synonym_name = self._get_name_from_synonym()
if synonym_name is not None:
  name = synonym_name
```
I could be talked out of it though - here for discussion...

fredkingham:
good point, changed

fredkingham:
You can now search by synonym, also changed it so that you can search by synonym in many to many although our front end doesn't allow it at the mo

davidmiller:
Searching for Diagnosis Condition contains knee doesn't return patients with diagnosis AKA for me on elCID 0.6.4 + Opal 1391-fix...

PR #1106, opened 2017-05-02
subrecord-display-hypothesis HYPOTHESIS, NOT FOR MERGING

fredkingham:
What do you think about this?

Essentially we want to decouple patient lists from get_display_template, or at least make it very flexible and easy for both patient detail and patient lists to use the same template.

In an ideal world, we would probably do something like

{% include models.Demographics.get_display_template(patient_list.get_prefixes) %}

and essentially this allows us to do this.

The in the patient detail views we can do the version of..

{% include models.Demographics.get_display_template(["walkin") %}

needs unit tests, its an experiment, thoughts?

### DIFF
--- a/opal/core/patient_lists.py
+++ b/opal/core/patient_lists.py
@@ -54,9 +54,7 @@ class ModelColumn(Column):
         self.single = model._is_singleton
         self.icon = getattr(model, '_icon', '')
         self.list_limit = getattr(model, '_list_limit', None)
-        self.template_path = model.get_display_template(
-            patient_list=self.patient_list()
-        )
+        self.template_path = ""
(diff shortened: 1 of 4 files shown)


davidmiller:
I like the general idea - that you pass in some extra prefixes to get_x_template...

It means we could do it for things that weren't just lists as well. e.g. custom display template for `Treatment` in a specific pathway.
I reckon you'd want to add it to e.g. `record_panel` as well ? 

For some reason this seems to back out some of the changes in https://github.com/openhealthcare/opal/commit/dfc1fa981ac0f87ae5104a4851002b9c3d77a33f but not others ?

fredkingham:
I think we all agree this is a good call

PR #1109, opened 2017-05-03
add requirements to the scaffold

fredkingham:


davidmiller:
We should definitely do this, but can we pin the Mock and Opal versions ?

fredkingham:
up for pinning mock, wasn't sure which version of opal to pin to, so the most recent in pypy seemed a good idea no?

davidmiller:
I think for now, do the same as the app `requirements.txt` - e.g. pin to a hard coded version.
(Maybe update the hard-coded 0.8.0 in the app requirements.txt)

But the right thing to do is to pin it to the version of Opal you used to run the scaffolding command with - e.g. the value of `from opal import __version__` and do it dynamically with requirements as a template.

davidmiller:
Great !

PR #1112, opened 2017-05-05
add a karma config to the plugin creation

fredkingham:
I always want this, but it does assume you're using the generic opal folder structure, ie that opal is ../../opal

I've put a comment, I think its good to have a karma config that will help you out of the box, my worry is if you don't have opal in the right place it may lead you astray. Thoughts?


fredkingham:
previous comment no longer applies, we now look at the location of opal in the scaffold, does kinda assume that that is the correct location, ie your running in your application venv or what have you, I think that's an ok assumption....

davidmiller:
But isn't that worse because it'll hardcode it to be the location on one specific machine? 

e.g. wouldn't it end up being `/Users/david/src/...` ?

PR #1122, opened 2017-05-15
card list template and unit tests added

fredkingham:
so we respect the patient list implementation schema which is pretty cool.

I purposefully haven't done the same thing with actions as I'm not 100% sure they are a function of the patient list and not a function of the episodes category/team


fredkingham:
take a peak at http://elcid-rfh-card-list.herokuapp.com/#/list/hepatology

PR #1127, opened 2017-05-16
Seperate out write to file

fredkingham:


opal/core/search/extract.py
```diff
@@ -263,37 +299,16 @@ def zip_archive(episodes, description, user):
...
-                )
-            else:
-                renderer = PatientSubrecordCsvRenderer(
-                    subrecord, episodes, user
-                )
-            if renderer.count():
-                renderer.write_to_file(full_file_name)
-                z.write(full_file_name, zip_relative_file_path(file_name))
+        generate_files(os.path.join(target_dir, zipfolder), episodes, user)
+        file_names = generate_files(root_dir, episodes, user)
```
davidmiller: I'm not sure I understand why we're calling `generate_files` twice?

fredkingham:
@davidmiller great call, done!

opal/core/search/extract.py
```diff
@@ -252,6 +252,42 @@ def __init__(self, model, episode_queryset, user, fields=None):
         )
 
 
+def generate_files(root_dir, episodes, user):
+    """ Generate the files and return a tuple of absolute_file_name, file_name
```
davidmiller: Referring to `the files` makes me think we're in a spy drama ;)

Can we call these e.g. `generate_full_extract_files` for more spelling but less likelihood of renaming later ?

fredkingham: do you say full extract because this will be in contrast to having inline extracts?

fredkingham:
@davidmiller what do you think of generate_csv_files?

PR #1129, opened 2017-05-18
refactored to use prefixes

fredkingham:


opal/models.py
```diff
@@ -54,6 +58,38 @@ def deserialize_date(value):
...
+        Subrecord.get_detail_template
+        Subrecord.get_form_template
+
+        into the new method of passing in a prefix.
+        It also warns the user about the deprecation
+    """
+
+    warnthem = """
+    opal.models.Subrecord.{} no longer takes
+    patient_list or episode_type.
```
davidmiller: Well, it does right... It _won't_ take them :)

I think all these messages should have a target version at which they will be removed?

opal/models.py
```diff
@@ -916,8 +952,29 @@ def get_display_name(cls):
         return cls._meta.verbose_name
 
     @classmethod
+    def _get_template(cls, template, prefixes=None):
+        result = []
```
davidmiller: Bikeshedding perhaps but... this isn't the `result` - I expect you to return it when I see this...

opal/models.py
```diff
@@ -940,61 +997,104 @@ def _build_template_selection(cls, episode_type=None, patient_list=None,
...
         """
         Return the active display template for our record
         """
-        templates = cls._build_template_selection(
-            episode_type=episode_type, patient_list=patient_list,
-            suffix='.html', prefix='records')
-        return find_template(templates)
+
+        if patient_list or episode_type:
+            prefixes = generate_prefixes_category_or_list(
```
davidmiller: This should probably extend `prefixes` ?

fredkingham:
@davidmiller I think these are fixed

fredkingham:
@davidmiller I think this is done

davidmiller:
Almost - I think there's no test for `_get_template` where prefixes isn't passed - see 

https://coveralls.io/builds/11604972/source?filename=opal%2Fmodels.py#L959

PR #1140, opened 2017-05-25
allows many to many fields to be from any lookup lists

fredkingham:
fixing the hard coded look up list for many to many


davidmiller:
This is plumbing for Advanced search M2M and doesn't change the interface, just generalises it?

Looks good...

Can we have an explicit test for the return `None` case of `get_lookuplist_api_name` ?

fredkingham:
This should have gone in, but the pypi release has gone now, shifting to 9 and will merge

PR #1145, opened 2017-05-26
Fix many to many select

fredkingham:


fredkingham:
This did go into v0.8.2 in the end

PR #1150, opened 2017-05-31
adds a warn template tag

fredkingham:


opal/templatetags/menus.py
```diff
@@ -17,7 +17,7 @@ def menu(context):
     """
     context = copy.copy(context)
     app = application.get_app()
-    menu = app.get_menu(user=context['user'])
+    menu = app.get_menu(user=context.get('user', None))
```
davidmiller: Is this part of the warnings PR ?

PR #1156, opened 2017-06-01
add elcid the elcid loggers into opal

fredkingham:


davidmiller:
@fredkingham I think this makes the tests quite noisy - see https://travis-ci.org/openhealthcare/opal/jobs/238320064

davidmiller:
Also: I think this probably wants to be not a sub point release... e.g. targets v0.8.3 or 0.9.0

TBH I can prob. live with waiting until 0.9 and porting the logger across  to e.g. RFH in the meantime...

Unless you particularly want to make an 8.3 release in the next week or so?

PR #1157, opened 2017-06-01
add options to the patient list as to whether we allow adding a patient or editing teams

fredkingham:


fredkingham:
Note I'm suggesting a merge into v0.8.2.2 as it fixes (by removing) the broken edit teams modals. 

It lets us declare edit teams and add new patient in the patient list.

opal/templates/partials/_teams_list.html
```diff
@@ -1,12 +0,0 @@
-<h4>
-  <i class="fa fa-users"></i> Teams
```
davidmiller: Isn't this exactly the case that we introduced the warning template tag for?

davidmiller:
Similar to the email logger one - can we target this at 8.3/9 ?

opal/templates/partials/_teams_list.html
fredkingham: yep, except that's in v0.9.0. We could leave it in just so that we can add in a warning for it in 0.9.1 but I'm not sure its worth it

PR #1175, opened 2017-06-08
remove the unused option to pass in a file name to load lookup lists

fredkingham:


davidmiller:
TBH I think that's precisely the wrong thing to do :) 

I think you should be able to tell this command to ignore it's 'be clever about where the files are'.

e.g. I work on two apps and I want to pass a relative path to the data from another app which I have on my system but it isn't in this app's `INSTALLED_APPS` yet...

I get what you mean about 'unused code', but I'd rather implement the option :)

fredkingham:
but at the moment its wrong! Implementing the option isn't hard, but I don't think we should have bits of code floating around that don't do anything that we'll probably implement in the future to make them work no?

PR #1177, opened 2017-06-11
Add eq/ne operators for the MenuItem class.

### LINKED ISSUE #1148: add equality to the menu items
add an __eq__ method to the MenuItem cls

pdmorrow:
This provides object equality for the menuitem class rather than "visual" equality  Unit test coverage remains at 100%.

Resolves #1148

### DIFF
--- a/opal/core/menus.py
+++ b/opal/core/menus.py
@@ -16,6 +16,23 @@ class MenuItem(object):
     def __repr__(self):
         return u"<Opal MenuItem href: '{0}'>".format(self.href)
 
+    def __eq__(self, other):
+        if not isinstance(other, MenuItem):
+            return NotImplemented
+
+        return (self.template_name == other.template_name and
+                self.activepattern == other.activepattern and
+                self.href == other.href and
+                self.icon == other.icon and
+                self.display == other.display and
+                self.index == other.index)
+
+    def __ne__(self, other):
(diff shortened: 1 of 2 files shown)


davidmiller:
This looks good to me - thanks @pdmorrow !

I suspect that I would _probably_ have split into two separate test methods for e.g. `==` and `!=` separately, but that's my taste more than anything particularly concrete :) 

@fredkingham - in the issue you [suggested](https://github.com/openhealthcare/opal/issues/1148#issuecomment-307644397) that 

> there's an argument that some attributes such as index don't need to be part of equality

FWIW I currently strongly feel that index is part of a sensible `__eq__` implementation and that it should operate on all of the properties. (As per this PR!)

Is there a particular use case I'm not thinking of?
I guess I might want to do some comparison based on the `href` but that's probably clearer written accessing the `.href` attribute directly?

fredkingham:
I'm just not sure you ever want for example to 'Add Patient' to appear in the menu items. That said I think its a non problem because as soon as you ran the application you'd see it and yeah maybe the best way to flag that wouldn't be in the __eq__ method anyway.

To conclude, looks great!

pdmorrow:
Thanks guys.  Sorry for the lack of replies during the day, day job is keeping me busy :)

PR #1178, opened 2017-06-11
Fix flake8 error.

pdmorrow:
7db573b Broke the flake8 run.


davidmiller:
Oh wow, great shout.

Thought my commit fixing this had gone into 0.9.0 but it's stuck on a feature branch - thanks @pdmorrow !

PR #1183, opened 2017-06-21
removing unused block

fredkingham:
I don't think this is ever used, it seems to come from 3 years ago from 

https://github.com/openhealthcare/opal/blame/v0.4.3/opal/static/js/opal/services/item.js

At that time I don't think we ever had integer fields, or if we did they were from observations and those aren't relative here.

I've tested with observations and time fields just to make sure and they work fine.


davidmiller:
It's about this I think: 

https://github.com/openhealthcare/elcid/commit/8d1b07630e9620f5ce92f830cbb1fe9560d60940#diff-cd78c938ec4beaa196abec0c07e0633bL348

PR #1186, opened 2017-06-25
moves the pagination of the fuzzy search into the database

fredkingham:


opal/core/search/tests/test_views.py
```diff
@@ -196,6 +197,29 @@ def test_incomplete_matching(self):
...
+        """
+        # we need to make sure we're all logged in before we start
+        self.assertIsNotNone(self.user)
+        for i in range(100):
+            self.create_patient(
+                "James", "Bond", str(i)
+            )
+
+        with self.assertNumQueries(35):
+            self.get_response('{}/?query=Bond'.format(self.url))
```
davidmiller: Why does searching for Bond do 35 database queries ? :)

fredkingham: I think that's because of the creation of patient summaries. Which is inefficient because we don't have a db only way to pull our episode start stop, so its manually goes through.

We could probably optimise it, but I don't think those 35 queries are causing us issues now. So they shouldn't cause issues in future. That's the theory anyway.

In fact this *should* (famous last words) be significantly faster, because this will mean that we do the pagination even in the database even when autocomplete search is disabled. So rather than do all patient summary aggregations on every request, we'll only get what we will display.

PR #1187, opened 2017-07-06
Bring in pathways

fredkingham:


davidmiller:
@fredkingham looks good ! 

2 things: 

- Js tests aren't passing :(
- The python tests here introduce an additional datetime warning - can we squash it by presumably `make_timezone_aware`ing the offending `datetime()` ?

davidmiller:
:boom:

PR #1199, opened 2017-07-19
removes discharge date and date of admission and date of episode adds…

fredkingham:
… in start and end


davidmiller:
@fredkingham This is great. So much cleaner with this refactor!

PR #1200, opened 2017-07-20
Force text on choices

fredkingham:


fredkingham:
we're going to do this in the serialiser instead

PR #1215, opened 2017-07-30
Cached modal template

fredkingham:


fredkingham:
Remove these from record panels as it makes the patient detail load too slow.
Leave in patient lists

PR #1218, opened 2017-07-31
the flow was not resolving

fredkingham:


fredkingham:
fixes openhealthcare/elcid#1437

davidmiller:
Having reviewed https://github.com/openhealthcare/opal/commit/5565d0354c9c9bbc4b485a62eabb5bc7de15dd88 I think it's obvious that I didn't really mean to delete the resolve line.

I guess it got caught in the diff with the whitespace shifts. This the real reason we should find a way to not be re-indenting javascript every other commit :)

PR #1219, opened 2017-07-31
Cached modal template

fredkingham:


opal/models.py
```diff
@@ -981,6 +981,14 @@ def get_form_url(cls):
         return reverse("form_view", kwargs=dict(model=cls.get_api_name()))
 
     @classmethod
+    def get_modal_url(cls, prefix=None):
+        url_kwargs = {}
```
davidmiller: Can we add reference documentation for this method ?

CHANGELOG.md
```diff
@@ -1,16 +1,27 @@
 ### 0.9.0 (Major Release)
 
+#### help text templates in select template tags are loaded in from the server
+This means they won't have access to local django context
+
+#### modals for subrecords in patient lists are cached on load
+These are now cached in an angular ng-template script tag in the page. This means that modals will load instantaniously for the user. It does mean that you can't use angular templates within your modals, we would recommend using client side rendering instead.
+
+#### cached_subrecord_modal
+This takes in a subrecord and caches its
```
davidmiller: This text doesn't actually tell you that it's a templatetag !

doc/docs/reference/panels_templatetags.md
```diff
@@ -57,3 +57,13 @@ Arguments:
 
 * `label`: The left hand item, to be rendered bold.
 * `model`: The right hand item, to be rendered as an angular expression
+
+#### {% cached_subrecord_modal subrecord, prefix(optional) %}
+
```
davidmiller: Can we add an HTML snippet that shows this in use?

opal/templates/_helpers/record_panel.html
```diff
@@ -1,3 +1,4 @@
+{% load panels %}
```
davidmiller: Why does this need to load panels now if nothing else has changed?

opal/templates/patient_lists/spreadsheet_list.html
```diff
@@ -220,6 +221,14 @@ <h4 class="print-only">
 
 {% block footer %}
 
+{% for column in columns %}
+  {% if column.model %}
+    {% for prefix in column.prefixes %}
```
davidmiller: Does that mean I include it multiple times if there are prefixes ?

Not sure why we need to loop over this if we then pass it in as an argument ?

opal/templatetags/panels.py
```diff
@@ -85,3 +85,34 @@ def aligned_pair(model=None, label=None):
...
+        if you put in a model and a patient list
+        it will do the reverse logic for you
+    """
+    child_context = copy.copy(context)
+    child_context.update(dict(
+        url=subrecord.get_modal_url(prefix),
+        template=subrecord.get_modal_template(prefix),
+
+        # we need to pass in the same context variables as
+        # opal.views.ModalTemplateView.get_context_data
```
davidmiller: This feels to me like it's going to break subtly when we add an extra context variable ?

PR #1224, opened 2017-08-06
a minor pathway refactor that changes goNext and goPrevious to use a …

fredkingham:
…history stack so that we can produce pathways that have conditional next steps in the future. Wel also add a way of updating the global editing dictionary. This is not documented because we're not using it in anger yet, its just internal changes


davidmiller:
This looks good but I think depends on subsequent PRs to be merge-able e.g. without the custom next steps it just adds ++ complexity :)

Issue #1117, opened 2017-05-10
As a user I want to be able to search by more than one team


fredkingham:
General queries often involve, I want all patients within 'tropical' and 'id inpatients' who have the diagnosis 'Lergy'

This will not be possible in our new and or or structure. Current theory is that what we want is the add many widget, but obviously it must be clear to the user what this means.

fredkingham:
This is significantly larger, we don't have access to the many to many select as no other query uses this. Symptoms just uses a a single select box, as you don't need to know if they have multiple selects.

Either we include the multi select or we go the custom template method

PR #1227, opened 2017-08-10
adds in multi select for tagging

fredkingham:


fredkingham:
fixes #1117

fredkingham:
So we now have team as a sub field of episode and with a custom search rule.

We also have a new filter type, which gives the select option a many to many select. its only used in this case but I think its useful new search rules.

I've added a new filter for display arrays in a more human readable fashion. E.g ["apples", "pears"] gets translated to apples and pears.

I've removed all mention of 'tagging' in the template.

opal/core/search/search_rule.py
```diff
@@ -109,7 +109,57 @@ def query(self, given_query):
             raise SearchException(err.format(given_query['queryType']))
 
 
+class EpisodeTeam(SearchRuleField):
+    ALL_OF = "All Of"
+    ANY_OF = "Any Of"
+
+    display_name = "Team"
+    description = "The team that looked after that episode of care"
```
davidmiller: I don't think that an episode of care is the subject of the verb in this sentence...

davidmiller: ...related, `the` is complicated b/c there can be more than one...

PR #1230, opened 2017-08-19
Extract form

fredkingham:


fredkingham:
This is now going into elcid first

PR #1242, opened 2017-08-21
fixes a bug where we are generating the summary based on patient id not episode id...

fredkingham:


davidmiller:
I've cherrypicked the first two commits into v0.9.0 so closing this one, the rest of this is noise I think?

PR #1256, opened 2017-08-27
Fix up episode unicode

fredkingham:


fredkingham:
```
except models.ObjectDoesNotExist:
             return self.start
```

fredkingham:
so, __unicode__ for Episode was broken, and then it turned out we didn't directly test these methods and they needed to be made to work with python 3 (a la https://docs.djangoproject.com/en/1.11/topics/python3/)

Also I conformed them all to work to the same pattern.

We can delete ContactNumber can't we?

fredkingham:
I think we're closing this for the moment

PR #1259, opened 2017-09-20
Every Module

davidmiller:
I read the Porting guide as suggesting that you should do this...

https://docs.djangoproject.com/en/1.11/topics/python3/


davidmiller:
This fails some tests that assert things are strings or serialize unicode naively :(

PR #1266, opened 2017-10-09
the test models should really inherit

fredkingham:


davidmiller:
@fredkingham when you say 'the patient lists should really inherit' do you mean 'the test models mimicing an application should really inherit' ?

fredkingham:
yep, edited accordingly

PR #1273, opened 2017-10-26
Minor Docs Update

JMathiszig-Lee:
Small improvement to the reference data documentation as I didn't quite understand what was meant by get_api_name() or where to find it easily. Some minor corrections for commands and the address for schemas


fredkingham:
Looks great!

davidmiller:
Thanks for the PR - It's really useful to know where these are confusing :) 

Minor note, I'm not sure the endpoint the schema docs are now pointing to is definitely correct - that's the extract schema, which is subtly different (in a way that should also probably be documented) and will probably move location relatively soon?

I think the endpoint we're after there is `/api/v0.1/record/` ?

JMathiszig-Lee:
I changed it to extract schema as /api/v0.1/schema/ doesn't appear in my list of API's so i thought they were the same thing. /api/v0.1/record/ certainly seems to accomplish the same thing but i didn't find it in the docs. Certainly no objection to any of that being changed though!

PR #1280, opened 2017-11-08
Use the plugin API api rather than viewset as view

davidmiller:


fredkingham:
I did not know this was a thing. Great call!

PR #1294, opened 2017-11-18
Update javascript indentation default for js-mode

davidmiller:
We should only have one, and be consistent...

If you're _very_ sure you want it to be 2, then this is the change that will update the Emacs default.

TBH I find it harder to read and crowded as compared to 4. Sure I could get used to it.

But pick one and stop moving whitespace back and forth would be good.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,2 +1,2 @@
 ((nil . ((tab-width . 4)
-         (js-indent-level . 4))))
+         (js-indent-level . 2))))


fredkingham:
honestly, I thought we were going with 2, if you find it more difficult to read, lets go with 4.

davidmiller:
> if you find it more difficult to read 
that could equally well be 'I'm less used to it'...

Your call @fredkingham If 2 then merge this, if 4 then find a way to reset Atom/Sublime/VSC  ? :)

Issue #402, opened 2016-01-13
Episode / Patient detail view for a patient that does not exist should give a nice error


davidmiller:
Currently we get an `alert()` - should be a nice 'not found' message and maybe even a search box

- [ ] Fix the display issues
- [ ] Note this in the changelog

davidmiller:
Closing given that the PR fixing this is merged.

PR #1295, opened 2017-11-18
402 patient detail 404

davidmiller:
All about having a friendly failure mode for the case when a user goes to a detail page that doesn't exist e.g. http://localhost:8000/#/patient/4009999994333333999292929 

(Refs #402 but doesn't close it because we're waiting on a merge to the branch that introduces the 9.1 Changelog section)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 ### Misc Changes
 
-Fixes some instances of progressbars not being set if unexpected error states occur.
+Fixes some instances of progressbars not being reset if unexpected error states
+occur.
+
+Improves the rendering of patient detail pages where no patient with the ID from
+route params exits. (Displays a polite message instead of erroring.)
 
 ### 0.9.0 (Major Release)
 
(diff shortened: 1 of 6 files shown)


fredkingham:
If we change this, we get an error message if the patient isn't found but not if the episode isn't found.

I think we merge this because its an obvious improvement and create another ticket?

Alternatives 

1. We use an http interceptor for all 404s and they send us to a 404 page by default
https://docs.angularjs.org/api/ng/service/$http#interceptors

The pros of an interceptor is it would work by default for wardrounds, pathways etc.

The cons would be that we'd be changing the page that would by weird in modals.

2. we do it in the service and redirect if it get's a 404 and we change all our services to use some sort of common, just go to the 404 page.

pros, easier to then implement else where
cons, we don't get a patient does not exist, we just get a does not exist (although its actually a different page so we could add context via the url), 

Other big cons for both of these is that they would break the back button, we'd need to do a location replace, but that's fine. Discuss.

davidmiller:
>wardrounds, pathways etc 

Right - this is small incremental improvement rather than deal with 404s etc at a framework level.

We already have a separate redirect to generic 404 system: https://github.com/openhealthcare/opal/blob/v0.9.0/opal/static/js/opal/controllers/patient_list.js#L16 but I'm far from convinced it actually works the way you'd want it to, hence making this localised incremental change here.

I'm not actually sure that a generic 404 behaviour pattern is even worth coding, testing, documenting...

PR #1311, opened 2017-12-04
set value of user to 'anonymous' if no user is passed. Test case exte…

### LINKED ISSUE #1304: Confidential emailer assumes that there is a user
In [opal/core/logConfidentialEmailer](https://github.com/openhealthcare/opal/blob/v0.9.0/opal/core/log.py) we assume that the request has a `user` attribute. This is actually only added by the Django middleware and if an exception is thrown before this middleware then all the things error.

```python
        if hasattr(record, "request") and record.request:
            if record.request.user.is_authenticated():
                user = record.request.user.username
            else:
                user = "anonymous"
```

### Success criteria

- [x] The value of the user variable should be `"anonymous"` if the request has no user attribute.
- [x] Add a  #[test](https://github.com/openhealthcare/opal/blob/v0.9.0/opal/tests/test_log.py) to check that this will be set appropriately

**Target Branch:** 0.9.1

pacharanero:
…nded to cover

Closes #1304

### DIFF
--- a/opal/core/log.py
+++ b/opal/core/log.py
@@ -19,11 +19,10 @@ class ConfidentialEmailer(AdminEmailHandler):
         detail = ""
         if hasattr(record, "request") and record.request:
             user = "anonymous"
-            
+
             if hasattr(record.request, "user"):
                 if record.request.user.is_authenticated():
                     user = record.request.user.username
-
 
             m = "Request to host {0} on application {1} from user {2} with {3}"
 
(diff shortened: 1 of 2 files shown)


opal/core/log.py
```diff
@@ -18,10 +18,12 @@ def emit(self, record):
         record.args = []
         detail = ""
         if hasattr(record, "request") and record.request:
-            if record.request.user.is_authenticated():
-                user = record.request.user.username
-            else:
-                user = "anonymous"
+            user = "anonymous"
+            
+            if hasattr(record.request, "user"):
```
davidmiller: Can we comment about the why - e.g. not blowing up if there's an error before the relevant middleware? 
I can entirely imagine me looking at this in 18 months and thinking this is superfluous...

Issue #312, opened 2015-10-04
use svg for the timeline linw


fredkingham:
we can be cleverer with the timeline line (ie the line between speech bubbles) so lets do that

PR #1317, opened 2017-12-05
add missing migration from previous PR

pacharanero:
this migration was missed from pull request https://github.com/openhealthcare/opal/pull/1312

### DIFF
--- a/opal/migrations/0033_auto_20171205_1553.py
+++ b/opal/migrations/0033_auto_20171205_1553.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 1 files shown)


davidmiller:
If #312 does require migrations at upgrade then we probably need that in a changelog note for the version like e.g. [this one](https://github.com/openhealthcare/opal/blob/v0.9.1/CHANGELOG.md#upgrade-instructions)
and also something in https://github.com/openhealthcare/opal/blob/v0.9.1/doc/docs/reference/upgrading.md presumably something like 'run makemigrations run migrate' ?

pacharanero:
This PR superseded by #1319 therefore closed

PR #1325, opened 2017-12-14
when deleting subrecords we will now set the created_by field to None

### LINKED ISSUE #1324: When I delete a user, please make sure it doesn't delete all related records
If a user creates a subrecord. At the moment when I delete that user, it deletes all related subrecords. 

Set on_delete=models.SET_NULL

fredkingham:
fixes #1324


davidmiller:
I think this is probably part of the right thing to do..
But we want more.

As a minimum I think we want to add an upgrade guide that tells you to run [make]migrations for your own application.

Honestly I think removing the delete user button in the Admin is probably the right thing to do if they have any subrecords.

Closing this in favour of #1329

PR #1327, opened 2017-12-14
when we delete a lookup list, let the fk_or_ft field handle it

fredkingham:


davidmiller:
This is great, but my feel would be that a minor point release isn't the place to change this behaviour...
Have merged into a 0.10.0 branch

PR #1334, opened 2017-12-29
Change javascript template model

fredkingham:


CHANGELOG.md
```diff
@@ -1,5 +1,97 @@
 ### 0.10.0 (Major Release)
 
+#### Subrecord variables in the forms
```
davidmiller: There are some typos/missing words in this changelog section...
I would probably move most of the detail here to the upgrade guide - I think that's where I want to know about the details of changes?

doc/docs/reference/upgrading.md
```diff
@@ -3,6 +3,15 @@
 This document provides instructions for specific steps required to upgrading your Opal
 application to a later version where there are extra steps required.
 
+### 9.1 -> 10.0
+
+the scope of variables in form has been changed from, for example editing.demographics.name to demographics.name.
```
davidmiller: Capitalisation, variable formatting etc  ! :)

opal/core/pathway/static/js/pathway/controllers/find_patient.js
```diff
@@ -25,10 +25,11 @@ angular.module('opal.controllers').controller('FindPatientCtrl',
 
     scope.new_patient = function(result){
         scope.state = 'editing_demographics';
+        scope.editing.demographics = [{}];
```
davidmiller: I don't think I follow why we need this as well as `scope.demographics` ? (Or indeed at all?)

opal/core/pathway/static/js/pathway/directives.js
```diff
@@ -1,171 +1,3 @@
-directives.directive("saveMultipleWrapper", function($parse){
```
davidmiller: This is an excellent file diff !

davidmiller:
Also, you may want to cherry-pick https://github.com/openhealthcare/opal/pull/1333/commits/83be00ecb42b6241b8181749e1dff5fe5a55ef3e (or review and merge that into 0.10 and then merge the branch) as that will reinstate coveralls checks.

davidmiller:
The more I think about this, the more I think that we still might not have it quite right.

We're trying to deal with the following things:

a) Dealing with the single instance `EditItem` world as well as the multiple instance Pathway world
b) Not being able to directly edit and then `save()` the `Item()` or `Episode()` properties
c) It feeling somehow awkward to edit model properties directly on `$scope` so you need something to hold a serialised `data` object.

Having e.g. `<div ng-repeat="condition in editing.condition">{% include models.Condition.get_form_template %}` is quite an awkward API to work with.

I would prefer say discouraging directly doing `{% include models.Condition.get_form_template %}` most of the time and instead providing template tags that do it for you should be a minimum.

I'm also not (yet) convinced that having an editing variable in Pathways but not in EditItemCtrl is the right way around to do this...

Is there not a way to have everything be closer to the JSON serialised episode? 
(I've not really spent much time with it recently, could just be hideously complex...)

e.g.
```
$scope.data = Episode().makeCopyWithSubrecords()
$scope.data = {item.api_name: [item.makeCopy()]}
```

What happens if we make the forms become equivalent to e.g. `model="record.provisional"` and then set record in wrapper templates for templatetags along the lines of these:

```
{% form_for models.Demographics %} <!-- Implicitly must be a singleton -->
{% form_for models.Diagnosis 0 %} <!-- Specify by index -->
{% multi_form_for models.Treatment %} <!-- render all of them that exist with add/remove ->
```

davidmiller:
`davidmiller closed this just now` <- Accidental - sorry !

fredkingham:
Delaying this for the time being

Issue #619, opened 2016-05-25
Episodes created via the gloss api have no created timestamp


fredkingham:

fredkingham:
closing this, I think its in error.

We should use the same system for creation and updating as subrecords. This should either be from update_from_dict or with django's auto_add_now

davidmiller:
> closing this, I think its in error.
Um, you haven't closed it though? :)

PR #1335, opened 2018-01-01
619 created timestamp in bulk update

davidmiller:


fredkingham:
I'm not sure this is right. I'm not sure we should be updating created without created by, we should also have an option for updated and updated by.

If we used update_from_dict whenever we updated episode both of these issues would be fixed. 

This is probably preferable but creates a lot more work.

davidmiller:
I don't think updated/created and updated_by/created_by are related.
I don't think I ever don't want created/updated set.
I'd be more open to considering it wrong because it should be auto_now / auto_now_add.

> I'm not sure we should be updating created without created by 
Why not?

> If we used update_from_dict whenever we updated episode both of these issues would be fixed.
I'm not sure I know what this means... 

This PR addresses this ticket: https://github.com/openhealthcare/opal/issues/619 
There are occasions when it might make sense to not have a user. 
There are no occasions when we don't want timestamps.

davidmiller:
@fredkingham If you've decided not to merge this then let's close it ?

fredkingham:
I'm closing this, given 
https://github.com/openhealthcare/opal/issues/1347 and  https://github.com/openhealthcare/opal/issues/1353

PR #1336, opened 2018-01-01
667 tagging updates

davidmiller:


opal/models.py
```diff
@@ -832,7 +843,7 @@ def get_tag_names(self, user, historic=False):
         if not historic:
             qs = qs.filter(archived=False)
 
-        return qs.values_list("value", flat=True)
```
fredkingham: we can just leave this as a values list can't we?

fredkingham:
Looks great! I should have done this ages ago.

Just not sure if we shouldn't convert the values list to lists in the tests rather than the otherway around. Ot

opal/models.py
davidmiller: It's quite deliberate... 

I think it's weird for set_tag_names to take a list of strings and get_tag_names to return a values list...

Made worse because those things print() identically.
You have to go and read the code to understand why `get_tag_names() != ['tag']` which is irritating.
Doubly so given that [the docs](http://opal.openhealthcare.org.uk/docs/reference/episode/#episodeget_tag_names) definitely imply it's a list of strings that comes back.

Why do you want it to be a values list?
There's no advantage I can see...

fredkingham: its just more efficient, you're pretty much always just going to iterate over it or serialise it, its only for equality that you wouldn't.

Microenhancements probably aren't necessary at this time so lets go with it.

PR #1344, opened 2018-01-02
copy is not needed here

fredkingham:


fredkingham:
This is not right, global context is still being updated.

PR #1346, opened 2018-01-04
adds in a unit test for create patient, removes a reference that pass…

fredkingham:
…es patient to create patient as this is unnecessary

### DIFF
--- a/opal/models.py
+++ b/opal/models.py
@@ -591,7 +591,7 @@ class Patient(models.Model):
             model = get_subrecord_from_api_name(api_name=api_name)
             if model in episode_subrecords():
                 if episode is None:
-                    episode = self.create_episode(patient=self)
+                    episode = self.create_episode()
                     episode.save()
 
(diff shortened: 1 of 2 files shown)


opal/models.py
```diff
@@ -591,7 +591,7 @@ def bulk_update(self, dict_of_list_of_upgrades, user,
             model = get_subrecord_from_api_name(api_name=api_name)
             if model in episode_subrecords():
                 if episode is None:
-                    episode = self.create_episode(patient=self)
+                    episode = self.create_episode()
```
davidmiller: Why are we removing this ?
Because it's implicit from `episode_set.create()`
Is there definitely a unittest that makes sure that bulk_update creates an episode for the correct patient?

fredkingham: our unit test for create_episode, explicitly checks this.

https://github.com/openhealthcare/opal/blob/v0.10.0/opal/tests/test_models.py#L204 also checks the behaviour of that additional episodes are created for the correct patient.

That test could potentially be improved though.

Issue #1279, opened 2017-10-27
consistency tokens are required in PUT requests, this is not clear


fredkingham:
you get an "Unexpected field name" which is hardly clear.

Steps to reproduce.

1. create a subrecord

2. update it without a consistency token e.g.

curl -X PUT http://127.0.0.1:8000/api/v0.1/past_medical_history/2/  -d '{"updated": null,  "created": "27/10/2017 12:40:00", "episode_id": 28, "updated_by_id": null, "details": "asdd", "year": "2017", "created_by_id": 1, "condition": "AKA"}' -H 'Content-Type: application/json' -H 'Authorization: Token 4fc401fc2e7508b455a6972917e2da6b1e4fbbc0' 

We get an opaque error message....

PR #1352, opened 2018-01-07
1279 required consistency tokens

davidmiller:
Fixes the issues noted in #1279


fredkingham:
looks great, discussion point, should delete's raise this exception if the consistency token has been raised elsewhere @davidmiller?

Theoretically yes, but I'd like your opinion.

I think that' would be a different ticket anyway.

Issue #1360, opened 2018-01-10
Get the testsuite working under Windows


lamby:
See, for example, #1297.

lamby:
https://www.appveyor.com/

lamby:
Filed: https://github.com/davidmiller/ffs/pull/13

lamby:
Filed: https://github.com/davidmiller/ffs/pull/14 — _Initial support for Windows, at least allowing module import_

lamby:
Filed: https://github.com/openhealthcare/opal/pull/1363 — _Add ghetto --failfast support to runtests.py_

lamby:
Filed https://github.com/davidmiller/ffs/pull/17 — _ffs.path: Correct tread -> treat typo_

lamby:
Filed https://github.com/davidmiller/ffs/pull/18 — _ffs.nix: Explicitly pass str instances not Path to shutil methods_

lamby:
Filed https://github.com/davidmiller/ffs/pull/19 — _Pass a genuine str instance to os.path.abspath, not our wrapper_

lamby:
![](https://i.imgur.com/ymtE0z1.jpg)

@davidmiller #1366 will make the testsuite work on Windows, given a recent enough ffs (ie. with my recent PRs merged, released and requirements.txt updated to match)

I've split off #1365 so that we can close this particular issue.

lamby:
@davidmiller Can you release a new ffs at let me know the version? :)

davidmiller:
Released and a version bump PR for you ^^

lamby:
:+1:

lamby:
Blocking on https://github.com/openhealthcare/opal/pull/1367#pullrequestreview-88964840

lamby:
Appveyor tests #1365 (passing!) are now also on this branch :)

lamby:
(ACK that tests are failing. Will look into soon.)

davidmiller:
(OPINION that the errors coming out of *nix are significantly more helpful due to providing 'context' and/or 'details')

Also: I didn't know `django.core.management.call_command` was a thing.
It seems ++better than subprocess presuming it can be made to work.

lamby:
Sure, I'm not stuck, just acknowledging they are failing..

davidmiller:
Closing this as we've been running on Appveyor for quite some time now...

lamby:
:+1:

PR #1363, opened 2018-01-11
Add ghetto --failfast support to runtests.py

lamby:


davidmiller:
Super - have been meaning to do this for ages ! :) 

Ideally, I suspect that:

(1) this option would also come through the `opal test` command which would abstract it for the opal project itself, concrete Django applications, and opal plugins ([like e.g.](https://github.com/openhealthcare/opal-wardround) ) 
Would need updating in:
- [Argparse instance](https://github.com/openhealthcare/opal/blob/v0.10.0/opal/core/commandline.py#L248)
- [Test function](https://github.com/openhealthcare/opal/blob/v0.10.0/opal/core/test_runner.py#L24)
- [Test function tests](https://github.com/openhealthcare/opal/blob/v0.10.0/opal/tests/test_core_test_runner.py)

(2) we would also have a way to make Karma respect failfast intentions for javascript testing.

Starting to sound like scope creep for #1360...

davidmiller:
Noted these comments/desires in #1364

PR #1381, opened 2018-01-17
Update Table list base template…

davidmiller:
…to understand the difference between patient_lists.Column and patient_lists.ModelColumn


fredkingham:
@davidmiller, great, but shouldn't this go straight to the episode detail page?

davidmiller:
Almost certainly - that's not changed though - feel free to update :)

Issue #1365, opened 2018-01-12
Add Windows tests on appveyor


lamby:
Split from #1360 

(https://www.appveyor.com/)

lamby:
You might receive some notifications whilst I get this working. Progress on the #1360 branch.

Issue #1361, opened 2018-01-10
Ensure that you can use the commandline tool in windows


davidmiller:
We should be able to run the four common `opal` commands on Windows - check that these work.

- [x] startproject
- [x] startplugin
- [x] scaffold
- [x] test

(Blocked by #1360)

Commands documentation: 
http://opal.openhealthcare.org.uk/docs/v0.9.0/guides/command_line_tool/

lamby:
``startproject`` works

lamby:
As in, ``startproject`` *now* works  :)

lamby:
These all work now, so closing issue. The code for them is on #1360 etc.

lamby:
@davidmiller On my windows, pip installs this script to ``C:\Python\Scripts\opal`` (or close enough!) that DTRT. This seems to be The Windows Way of doing it.

davidmiller:
On mine too although that script doesn't seem to be editable...

e.g. when I run the script from powershell, it opens the file in notepad instead of running it... I guess my immediate question is what are we doing differently?
![screen shot 2018-01-25 at 14 01 42](https://user-images.githubusercontent.com/17229/35392164-e7a8c694-01d8-11e8-9989-d07ada1ad518.png)

lamby:
``C:\Python27\python.exe bin/opal startproject`` :)

davidmiller:
Ah, I see...

I think that's much less good than converting `scripts` to `entry_points` which creates an `opal.exe` so that inside a virtualenv you can type e.g. :
`opal startproject` 

Our documentation is going to be much less confusing for all concerned that way :) 

To learn this I had to do the work - stacked branch is [here](https://github.com/openhealthcare/opal/commit/d997fb8af5198ef9e4657faf51f70ebf8322f3ff) - TBH we can merge that after PR #1385

davidmiller:
Definitely seen this working now - closing.

lamby:
:+

PR #1385, opened 2018-01-20
Fix Windows testsuite and main management comments

lamby:
This resolves issues:

 * #1360 - Get the testsuite working on Window
 * #1365 - Test opal on AppVeyor (Windows)
 * #1361 - Ensure we can run the main management commands on Windows.


opal/tests/test_core_commandline.py
```diff
@@ -158,7 +158,8 @@ def test_test(self):
         with patch.object(commandline.test_runner, 'run_tests') as rt:
             commandline.test(mock_args)
             mock_args_call = rt.call_args[0][0]
-            self.assertTrue(mock_args_call.userland_here.endswith("opal"))
+            self.assertTrue(mock_args_call.userland_here.endswith("opal"),
+                            "Not running tests from root directory")
```
davidmiller: Nice.
I have never read the call signature of `assertTrue` closely enough to notice the second message.

davidmiller:
This looks great, and demonstrably passes the tests :)

Can you talk me through the thinking behind https://github.com/openhealthcare/opal/commit/e918a6a0a6b110e0829127305a709790299b7ceb ? 

e.g. I can read what you're doing, but would like to understand why you're doing it...

opal/tests/test_core_commandline.py
lamby: All the assertFoo calls take messages. Getting them phrased gramatically is a bit of PITA though as they "look" like comments in the code but aren't displayed in such a context when being run. :)

lamby:
Re. e918a6a shall I comment there or here?

davidmiller:
Wherever you like - I think I get notifications from everywhere ...

lamby:
@davidmiller 

So the issue here was that Django doesn't *quite* support completely reloading ``django.conf.settings``. It somewhat pretends to, in that it can temporarily override settings and suchforth, but every attempt I have ever made at "please **move** to these settings over here" (ie. what we want to do here) doesn't work within the same Python interpreter. :)

The problems are mostly around things being imported twice (or other import side effects), but there are also problems about defaults - ie. if the current settings override a default setting but the "new" 


We basically bypass this by using a new global Python interpreter - we already are calling `manage.py` for a bunch of other things anyway so it would be no less ugly to do this and indeed, it makes might sense to do it like this anyway even if we were not having any settings issues atm. :)

opal/scaffolding/plugin_scaffold/app/plugin.py.jinja2
```diff
@@ -13,9 +15,9 @@ class {{ name.title() }}Plugin(plugins.OpalPlugin):
     javascripts = {
         # Add your javascripts here!
         'opal.{{ name }}': [
-            # 'js/{{ name }}/app.js',
```
fredkingham: Is this right? I might be wrong but I think this should stay like it was

lamby: (on mobile) Might be a preprocessor/concatenator argument...

fredkingham: for context it get's passed to https://github.com/openhealthcare/opal/blob/v0.10.0/opal/templates/plugins/javascripts.html#L3

davidmiller: These strings end up passed to 
```
# ./opal/templates/plugins/javascripts.html
<script type="text/javascript" src="{% static variable %}"></script>
```
Does `{% static %}` take url paths or file paths ? I've never had to know... 
Looks like it's a URI to me?

lamby: Fixed in https://github.com/openhealthcare/opal/commit/97d2b94672c3b959e2d2dd4941c66e99825dee19 :)

PR #1389, opened 2018-01-26
fixes deletion of lookup lists on abstract models

fredkingham:
the on_delete needs manage abstract and proxy classes


fredkingham:
the on_delete needs manage abstract and proxy classes

PR #1395, opened 2018-02-01
If we use entry_points rather than scripts to make $ opal it works ni…

davidmiller:
…cerer on windows...


fredkingham:
That's pretty neat!

PR #1402, opened 2018-02-08
353 pathway form size

fredkingham:
so the issue in openhealthcare/elcid-rfh#353 was actually caused because we artificially inflate the modal form size which I didn't realise.

Text/selects/inputs are size 18, so now page pathways do the same (modal pathways already did).

I've also made the same true for the log in screen.


fredkingham:
fixes openhealthcare/elcid-rfh#353

PR #1405, opened 2018-02-09
Make axes version in scaffold requirements.txt match that in setup.py

davidmiller:


pacharanero:
Is there any way that process could be automated?
eg the jinja2 template would pull in the information from the relevant bit of setup.py? 
Or are there instances where you would not want that behaviour?

davidmiller:
> eg the jinja2 template would pull in the information from the relevant bit of setup.py?

I think that's a very decent shout actually.
(Unless I'm missing some subtle reason why it's hard...)

Paste from repl:
```python
>>> from pip._vendor import pkg_resources
>>> package = pkg_resources.working_set.by_key['opal']
>>> [str(r) for r in package.requires()]
['ffs>=0.0.8.2', 'letter==0.4.1', 'jinja2==2.10', 'requests==2.18.4', 'django==1.10.8', 'django-reversion==1.10.2', 'django-axes==1.7.0', 'djangorestframework==3.4.7', 'django-compressor==1.5', 'python-dateutil==2.4.2', 'django-celery==3.1.17', 'celery==3.1.19', 'six>=1.10.0']
```

PR #1406, opened 2018-02-12
Scaffold fix up

fredkingham:


fredkingham:
@davidmiller tested on windows (python 3.6.4), ubuntu (python 2.7) and tested a mac on python 2.7 and 3.4.2 all works fine. I *think* this is enough.

PR #1384, opened 2018-01-19
defers the setting of episode stage to the category

fredkingham:

PR #1407, opened 2018-02-13
Store stage in the db

fredkingham:
depends on #1384

stores start/stop for stages.


opal/core/episodes.py
```diff
@@ -63,20 +65,36 @@ def has_stage(self, stage):
...
+
+        current_stage = self.episode.stage_set.filter(stopped=None).last()
+        now = timezone.now()
+        if current_stage:
+            current_stage.stopped = now
+            current_stage.updated = now
+            current_stage.updated_by = user
+            current_stage.save()
+
+        if stage_value:
```
davidmiller: What's the point of this `if` ? 
Isn't it redundant b/c the `.has_stage()` call above?

opal/models.py
```diff
@@ -1684,6 +1707,19 @@ def explicit_access_only(self):
...
+        one with no stop date any time.
+
+        This is expected to be set by the episode_category
+        which determines what stages are available to an episode.
+    """
+    episode = models.ForeignKey(Episode, null=False)
+    started = models.DateTimeField()
+    stopped = models.DateTimeField(blank=True, null=True)
+    value = models.CharField(max_length=256)
+
```
davidmiller: Is it possible to have e.g. episode and stopped unique_together? 
Will this prevent two stages for a single episode without an end date?

opal/models.py
```diff
@@ -754,6 +754,12 @@ def category(self):
         )[0]
         return category(self)
 
+    @property
+    def stage(self):
+        s = self.stage_set.filter(stopped=None).last()
```
davidmiller: I feel like I want this to be `.get()` and error if we can't set it as a constraint on the model...

opal/models.py
```diff
@@ -865,6 +871,15 @@ def _episode_history_to_dict(self, user):
         episode_history = episodes_for_user(episode_history, user)
         return [e.to_dict(user, shallow=True) for e in episode_history]
 
+    def update_from_dict(self, data, user, force=False, fields=None):
+        # stage is a related model so episode
+        # needs to have been saved before we can
```
davidmiller: This comments leave out the last few words that would make it a complete

davidmiller:
I notice that there are no docs for this at all...

opal/core/episodes.py
```diff
@@ -63,20 +65,36 @@ def has_stage(self, stage):
...
+        if not self.has_stage(stage_value):
+            if stage_value is not None:
                 msg = "Can't set stage to {0} for {1} Episode".format(
-                    stage, self.display_name
+                    stage_value, self.display_name
                 )
                 raise ValueError(msg)
-        self.episode.stage = stage
+
+        current_stage = self.episode.stage_set.filter(stopped=None).last()
```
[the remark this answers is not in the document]
fredkingham: I agree, its just we don't usually do blowing up. 

We can put in a `get` but really we don't want a get blowing up here as we'd want it to blow up on `save` of the flawed record.

IE, if we want to change this to a `get` we should add another check at another point. Its usually our MO that we don't blow up. Do you think we should?

opal/core/episodes.py
fredkingham: 👍

fredkingham: Actually this is maybe a different bug. 

Presumably our api for stopping the stage is to set stage to null on the client.

So we should handle this appropriately

opal/models.py
fredkingham: so we use last here because stage it not necessary set so I think this is ok.

fredkingham:
@davidmiller so a bunch of changes.

Episodes can have 'no' stages and stages need to be explicitly set.

This is so we can work through different episode flows and sort them out individually.

Instead the current restrictions have been put in place.

1. You cannot save an episode stage if another is already open.
2. You cannot set an episode stage to None if the episode set has been previously set.

Change log and documentation have been updated accordingly. Let me know what you think.

opal/models.py
```diff
@@ -1684,6 +1701,35 @@ def explicit_access_only(self):
...
+    @transaction.atomic
+    def save(self, *args, **kwargs):
+        """ An episode should not have multiple open stages.
+            raise exceptions and roll back if anyone tries
+            any funny business.
+        """
+        super(Stage, self).save(*args, **kwargs)
+        existing = Stage.objects.filter(episode=self.episode)
+        existing = existing.filter(stopped=None)
+        if existing.count() > 1:
```
davidmiller: You're relying on transaction.atomic rolling it back and not saving ?
Does a unique_together constraint not work?

fredkingham: so we can do a unique_together in theory but its a bit of a fudge.

What we want 
```
   unique_together = (('episode', 'value', 'stopped'),)
```
but really we 'd like
```
   unique_together = (('episode', 'value', 'stopped' != None),)
```
Although as they are datetime field they are probably the same thing... I thought this was was a bit nicer as its an explicit check

fredkingham:
moving to v0.10.1

PR #1408, opened 2018-02-14
Sometimes people make their Django models a package. Opal scaffolding…

davidmiller:
… should support that use case.


fredkingham:
looks great, tested and confirmed it works.

PR #1413, opened 2018-02-17
we're missing a comma in our text area height declaration in the css

fredkingham:


davidmiller:
@fredkingham is that what's making the height seem weird?

PR #1415, opened 2018-02-21
Port the detail controller .refresh() function to the list controller

davidmiller:


opal/static/js/test/patient_list.controller.test.js
```diff
@@ -207,8 +207,30 @@ describe('PatientListCtrl', function() {
 
     });
 
+  describe('refresh()', function() {
+      fit('should update .episodes with what the server thinks is the episode', function() {
```
fredkingham: witness the fitness?

(just it)

davidmiller: Good shout - and you got there first with the fix too :)

pacharanero:
are we happy to merge so I can test downstream epidur.io foo?

pacharanero:
tested locally with the use case in Epidur.io, all working

PR #1417, opened 2018-02-24
ignore abstract lookup lists

fredkingham:


fredkingham:
@davidmiller is this good to go?

PR #1418, opened 2018-02-28
work in progress improvement to the theming documentation

pacharanero:
* used {{yourapp}} as 'placeholder' name throughout
* added some basic theming info
* have not yet approached including CSS in Opal scaffolding but maybe this is the way forward


davidmiller:
Nice !

PR #1420, opened 2018-03-01
I *think* the missing word was meant to be 'singleton'

pacharanero:
If so please find attached a PR fixing the missing word and with a link to the relevant bit of the doco

### DIFF
--- a/doc/docs/tutorials/pathway_tutorial.md
+++ b/doc/docs/tutorials/pathway_tutorial.md
@@ -68,7 +68,7 @@ class SimplePathway(pathway.PagePathway):
     )
 ```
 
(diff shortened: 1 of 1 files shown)


davidmiller:
I think so to! 

However: 
- You've typo'd singleton the second time
-  links should be relative, pointing at markdown files rather than to a specific version of the docs I reckon ?

pacharanero:
good points, well made
I've fixed the typo

Any thought on how best to create a relative link that works for both the compiled docs and the github markdown versions?

davidmiller:
Like this http://www.mkdocs.org/user-guide/writing-your-docs/#linking-documents 
(relative path to the .md file you're targetting)

pacharanero:
Thanks
I've updated the PR accordingly. I haven't attempted to build the docs locally though

davidmiller:
Checked it and it links correctly...

PR #1424, opened 2018-03-02
pointer to Guides section added.

pacharanero:
To me I would have thought the most detailed section of doco would be the Reference section, so if that's not the case (as in here where the're more detail in the Guides section on Steps) then we need to give a pointer to where the the detail is


davidmiller:
The Tutorials / Guides / Reference split should be broadly as described here: https://jacobian.org/writing/what-to-write/ IMHO :)

PR #1429, opened 2018-03-02
undefineds are not serialised, so we explicitly set start end to null

fredkingham:
fixes openhealthcare/elcid#1545 

Undischarge date of discharge was not being set because we were setting end to `undefined` rather than `null`.

 `undefined` is not serialised to json, we weren't removing it when it hit the server.

### DIFF
--- a/opal/static/js/opal/services/episode.js
+++ b/opal/static/js/opal/services/episode.js
@@ -155,11 +155,23 @@ angular.module('opal.services')
 
             makeCopy: function(){
                 var start, end;
+                /*
+                * undefined is not serailised.
+                * we explicitly set to null, to set
+                * fields that were populated to null
+                */
                 if(this.start){
                   start = moment(this.start).toDate();
                 }
+                else if("start" in this){
+                  start = null;
+                }
+
                 if(this.end){
                   end = moment(this.end).toDate();
+                }
+                else if("end" in this){
+                  end = null;
                 }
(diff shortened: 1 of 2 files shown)


opal/static/js/opal/services/episode.js
```diff
@@ -155,12 +155,24 @@ angular.module('opal.services')
...
                 var start, end;
+                /*
+                * undefined is not serailised.
+                * we explicitly set to null, to set
+                * fields that were populated to null
+                */
                 if(this.start){
                   start = moment(this.start).toDate();
                 }
+                else if("start" in this){
```
davidmiller: What's the point of these `in` checks ?
How would they fail?
It's an episode, don't we know from first principles that it always has a start and an end ?

fredkingham: so the reasoning for this is because update from dict will not update properties that are not explicitly in the dict. Therefore if you wanted to just do a parital update you would want to make copy something that had an end for example without mentioning a start.

That said... We don't actually use this use case with our js code and as you say an episode should always have a start to the extent it should probs be a required field that defaults to today.

fredkingham: I've still done the explicit set start to null if its not set as we shouldn't do start date validation in js if we were going to do it, and I think this is the more opally expected behaviour.

PR #1432, opened 2018-03-03
adds in rhino virus to the investigation

fredkingham:
fixes https://github.com/openhealthcare/elcid/issues/1544


davidmiller:
@fredkingham shouldn't there be a migration ?

fredkingham:
@davidmiller its an abstract model, so migrations would be in whatever inherits from it, so there will be a corresponding migration in elcid required. I was going to wait until this was merged and just commit the migration directly.

davidmiller:
Merged outside this PR

PR #1440, opened 2018-03-08
Split settings up into Django, Third Party, Opal, and Project blocks

fredkingham:


opal/scaffolding/scaffold/app/settings.py.jinja2
```diff
@@ -150,108 +78,154 @@ TEMPLATES = [
...
+            'HOST': '',
+            'PORT': ''
         }
+    }
+
+# Password validation
+# https://docs.djangoproject.com/en/{{ docs_version }}/ref/settings/#auth-password-validators
+AUTH_PASSWORD_VALIDATORS = [
+    {
+        'NAME': 'django.contrib.auth.password_validation.UserAttributeSimilarityValidator',
```
fredkingham: can we double check that this doesn't break

https://github.com/openhealthcare/opal/blob/84b809a4a29b84e97730018de9e49517f2564ef8/opal/scaffolding/scaffold/app/management/commands/createopalsuperuser.py.jinja2

It shouldn't because the docs say they get validated in the form.

In dango 2 though some validation also get's applied on the command line so can we check...

fredkingham:
this is old news now and has been superseded by 
https://github.com/openhealthcare/opal/pull/1444

PR #1443, opened 2018-03-09
fix to default tests.py in the Opal scaffold

pacharanero:
changed `name` to `first_name` as per change to models https://github.com/openhealthcare/opal/commit/b318d4954f18ba60677b99dc2bdc6e6aaac25c1c


fredkingham:
👍

PR #1444, opened 2018-03-09
Update Settings Scaffold

ghickman:
Base the Django settings scaffold file off Django 1.10's template since it's pinned in requirements.txt. Split the settings up into sections.


fredkingham:
looks great

PR #1445, opened 2018-03-09
Document and Script Releases

ghickman:
Document how to make a release and wrap the PyPI steps up in a makefile to make `make release` possible.

I've added `make docs` as a shortcut for running the documentation engine too.


doc/docs/guides/releasing.md
```diff
@@ -0,0 +1,6 @@
+# Making a Release
+1. Tag the HEAD of the release branch and push to github: `git tag <version> && git push --tags`
+
+1. Check version is correct in `_version.py`.
+
+1. Make sure you have `twine` installed and then: `make release`
```
fredkingham: @ghickman  why not add twine to the requirements?

ghickman: I considered that but it's only necessary for people who are uploading releases to PyPI so seemed superfluous for typical development.  Your call though!

Makefile
```diff
@@ -0,0 +1,15 @@
...
+	@echo "    make docs    -- setup and serve docs locally"
+	@echo "    make release -- build and release to PyPI"
+
+docs:
+	pip install -r doc/requirements.txt
+	mkdocs serve
+
+release:
+	rm -rf dist/*
+	python setup.py register bdist_wheel sdist
```
davidmiller: Why register here ? I thought you only need to do that once ?

doc/docs/guides/releasing.md
```diff
@@ -0,0 +1,6 @@
+# Making a Release
+1. Tag the HEAD of the release branch and push to github: `git tag <version> && git push --tags`
+
+1. Check version is correct in `_version.py`.
```
davidmiller: Are you deliberately writing a numbered list with all the numbers as 1. 1. 1. ?

davidmiller:
Mostly this looks great - couple comments and I've added some further steps in #1450

doc/docs/guides/releasing.md
davidmiller: I think this is fine tbh - let's see if it actually trips anyone up :)

doc/docs/guides/releasing.md
pacharanero: IIUC this is correct Markdown, any markdown preprocessor will auto-renumber a 1. 1. 1. 1. list as 1. 2. 3. 4, which makes it easier if you needed to re-order the items in the list ever.

I guess the downside is that anyone viewing the unpreprocessed source will not see a correctly numbered list.

ghickman: @pacharanero is correct here, it's how one does a markdown list for auto-numbering.  It's definitely a bit odd at first view but if you ever have to insert numbers into a list it's much easier!

Makefile
ghickman: Register sets metadata on the PyPI so if you ever change something like Trove classifiers or author email in your setup.py it wouldn't be updated in the PyPI until you run `register` again.

Doing it on every release is probably overkill but it removes you ever having to think about it and is harmless when nothing has changed.

davidmiller: I did not know that upload wouldn't update metadata !

PR #1446, opened 2018-03-09
Attempt to redirect the user to where they wanted to go on CSRF failure

### LINKED ISSUE #596: stale csrf token
can we fix it so that if you haven't logged in for a while it doesn't give you the csrf issue

ghickman:
This addresses the case where an unauthenticated user ends up with two
browser windows/tabs pointing at the login form.  Both versions of the
page will have valid CSRF tokens in their forms.  The user then logs in
with one of those pages and Django cycles the CSRF token, making the
second page's token invalid.  The user then tries to log in via the
second page and gets a CSRF failure/403.

This change redirects the user to their intended page via the `next`
param (falling back to /).  An unauthenticated user should be redirected
to the login page again (getting a valid CSRF token).

Fixes #596

### DIFF
--- a/opal/views.py
+++ b/opal/views.py
@@ -4,7 +4,7 @@ Module entrypoint for core Opal views
 from django.core.urlresolvers import reverse
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib.auth.views import login
-from django.http import HttpResponseNotFound
+from django.http import HttpResponseForbidden, HttpResponseNotFound
 from django.shortcuts import get_object_or_404, redirect
 from django.template.loader import get_template
 from django.template import TemplateDoesNotExist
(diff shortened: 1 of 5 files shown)


fredkingham:
@ghickman looks great, can we have a unit test please?

(opal is 100% unit tested)

davidmiller:
For me, I don't particularly think there's any value in our testing the [Django setting](https://docs.djangoproject.com/en/1.10/ref/settings/#csrf-failure-view).

I'd unittest our view function with mock requests and that'd be enough... (e.g. wouldn't bother with GET portion of the flow).

We'll also want to explain that upgrading applications will want to add this to their settings.py in the [upgrade guide](https://github.com/openhealthcare/opal/blob/v0.10.0/doc/docs/reference/upgrading.md#091---0100) and note the addition in the Changelog [misc changes section](https://github.com/openhealthcare/opal/blob/v0.10.0/CHANGELOG.md#misc-changes)

opal/views.py
```diff
@@ -299,7 +299,7 @@ def get(self, *args, **kw):
 
 def csrf_failure(request, reason):
     if request.POST:
-        next_url = request.POST.get('next', '/')
+        next_url = request.GET.get('next', '/')
```
davidmiller: I completely missed this :)

ghickman: Me too!  Good job @fredkingham made me add some tests :P

PR #1449, opened 2018-03-09
the version of django-celery doesn't work with django 1.10.8

fredkingham:


davidmiller:
@fredkingham Needs to be in for instance: Changelog, upgrade guide, scaffold...

Also, what does the error look like, and is it possible to write a test to stop this from happening again ?

PR #1451, opened 2018-03-14
Experimental Import/Export Feature

ghickman:
I wasn't sure if you want this merged but felt it was best to have some record of discussion!

So, this is fairly rough first pass at Import/Export and Patient and Episode data via the Admin and a very basic form.

Some things I found along the way:
* Exporting needs to mutate the data slightly, removing IDs and adding some meta fields the importer can use to identify this as external data.
* You've tackled similar problems elsewhere with the API et al so at least exporting the data isn't _that_ hard.

Other points of interest:
Building up the correct models to import was done very manually because I didn't want to dive into a more generic solution for a proof of concept, however I think it's a very doable feature.

The code has become very generic between the pairs of views and is likely ripe for a properly generic view at some later stage.


davidmiller:
Closing this for now - it needs work still, and is pointing at the wrong branch.

PR #1460, opened 2018-03-21
change to use the getChoices function

### LINKED ISSUE #1459: Search for many to many does not work
To reproduce
- go to an instance of opal that has a many to many search for example elcid walkin symptoms
- go to extract
- select symptoms then symptom

the box looks like this

screen shot 2018-03-21 at 13 32 07

fredkingham:
fixes #1459 

Now looks like

screen shot 2018-03-21 at 13 34 38


fredkingham:
(also works with synonyms natch)

PR #1465, opened 2018-03-21
warning to those unwise hyphen-loving souls who would otherwise have …

pacharanero:
…to spend an hour+ debugging a missing template, because SLUGS DON'T LIKE HYPHENS

https://github.com/openhealthcare/opal/blob/v0.10.0/opal/core/pathway/urls.py#L8


davidmiller:
That seems suboptimal :(

Closing this unmerged to fix with #1467 instead.

PR #1467, opened 2018-03-22
Pathway slug hyphens

davidmiller:
I can't think of a good reason we don't allow pathway slugs to include `-` characters ?


pacharanero:
LGTM, this is a better solution than putting 'no-hyphens' in the documentation

davidmiller:
Agree. Not to mention the fact that it would have been adding insult to injury to hyphenate 'no-hyphens' in the documentation when doing so.

PR #1472, opened 2018-03-26
big refactor of the PatientList reference documentation including mor…

pacharanero:
…e up to date information about the available templates, documents the actual syntax for changing template_name, adds in all the missing properties thqat are mentioned in Guides but not in Reference, and puts properties in aplhabetical order within the class

#EAFP


fredkingham:
I've added additional opinionated changes, so I'm going to pass this over to @davidmiller for a final review

doc/docs/reference/patient_list.md
```diff
@@ -1,10 +1,9 @@
-# `opal.core.patient_lists`
+## opal.core.patient_lists
 
 The `patient_lists` module defines a number of classes for working with lists of patients.
 
-## `class  PatientList`
 
-### Properties and Class Methods
+### class opal.core.patient_lists.PatientList
```
pacharanero: if we're doing this: `opal.core.patient_lists.PatientList`, shouldn't we do the same in `TabbedPatientListGroup` and `TaggedPatientList`?

Is it necessary though? it says `opal.core.patient_lists` at the top of the page.

PR #1473, opened 2018-03-26
Elcid 1557 do not show discharge episode on mine

pacharanero:
* creates Django `{% block sidebar %}` and `{% block episode_actions %}`
* this is an upstream change to Opal that was required for Elcid-UCH issue 1557 https://github.com/openhealthcare/elcid/issues/1557


fredkingham:
@pacharanero :+1 

Looks great

PR #1476, opened 2018-03-27
1443 can we stop double clicks in the spread sheet view

pacharanero:
Adds an underscore.js `debounce` wrapper around `self.openEditItemModal` returning `self.debouncedOpenEditItemModal`.

Calls to this function are debounced so they can only be triggered once in each debounceTime (I've set to 200ms for now). The last parameter `true` is the 'immediate' flag to tell underscore to trigger on the _leading edge_ of the event (trigger **then** lockout for 200ms), as opposed to triggering at the end of the 200ms (lockout for 200ms **then** trigger).

I've tested this in elcid-uch both manually (trying to double click and trigger two modals) and also by putting two sequential calls to `self.openEditItemModal` in the `self.editItem` function.

With two sequential calls to `self.openEditItemModal`, two modals open.
With two sequential calls to `self.debouncedOpenEditItemModal`, only one modal opens.

opal branch for testing: 1443-can-we-stop-double-clicks-in-the-spread-sheet-view

closes openhealthcare/elcid#1443

### DIFF
--- a/opal/static/js/opal/services/record_editor.js
+++ b/opal/static/js/opal/services/record_editor.js
@@ -47,7 +47,11 @@ angular.module('opal.services').factory('RecordEditor', function(
       }
     };
 
+
     self.openEditItemModal = function(item, name){
+
+      console.log("modal" + Date.now() );
+
       $rootScope.state = 'modal';
       var template_url = '/templates/modals/' + name + '.html/';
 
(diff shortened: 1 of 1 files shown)


fredkingham:
@pacharanero, you have no idea how happy this makes me. It works perfectly but it needs unit tests

fredkingham:
closing and moving to v0.10.1

PR #1479, opened 2018-03-30
As menuitem

davidmiller:
Enable the `x.as_menuitem` pattern for Patient lists and pathways.

Add basic `get_absolute_url` implementation to patient lists and pathways.

Adds a `get()` utility function to look for getters.


fredkingham:
so we definitely need this...

A couple of questions on the implementation. The get() method. Do we need this? It seems we are using it instead of properties so that we can use if for classmethods?

It feels unpythonic. I know pathways does do it a lot though but it we don't seem to be using it in this code unless I missed something.

get_absolute_url for pathways needs to take args/kwargs right? Then pass them straight through to the format string?

fredkingham:
(moving to 0.10.1)

PR #1484, opened 2018-04-04
As menuitem

fredkingham:
see https://github.com/openhealthcare/opal/pull/1479 for details


davidmiller:
Let's move this to 0.11

PR #1486, opened 2018-04-04
Store episode in the db

fredkingham:
moving to v0.10.1 old pull request #1407


davidmiller:
This now needs to be re-pointed ? 

I _think_ the models conflict is because we deleted at least one of those methods?

davidmiller:
@fredkingham Can we check that you still think this is good to go then point at a better branch?

fredkingham:
repointing

PR #1487, opened 2018-04-05
by changing the order around we register urls, we can register urls i…

fredkingham:
…n the plugins that will override the core apis

### DIFF
--- a/opal/core/api.py
+++ b/opal/core/api.py
@@ -390,32 +390,36 @@ class PatientListViewSet(LoginRequiredViewset):
         return json_response(patientlist.to_dict(request.user))
 
 
-router.register('patient', PatientViewSet)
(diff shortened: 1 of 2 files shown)


opal/core/api.py
```diff
@@ -390,32 +390,36 @@ def retrieve(self, request, pk=None):
...
 
-    router.register(sub_name, SubViewSet)
+        router.register(sub_name, SubViewSet)
 
 
-def register_plugin_apis():
-    for plugin in plugins.OpalPlugin.list():
-        for api in plugin.get_apis():
-            router.register(*api)
+def initialise_router():
```
davidmiller: 14 instances of initialise and 18 instances of initialize in our codebase. 
We should pick one.

PR #1492, opened 2018-04-06
379 page loads behind proxy

fredkingham:
All fonts are now locally sourced.
Instead of having multiple style sheets, we move them into one and use media queries.
Google analytics is moved to the bottom of the file. This has been tested and still works.


fredkingham:
Also confirmed that sites behind a proxy now load as fast as you would expect. The apps now load quite a bit faster locally with dev tools open.

pacharanero:
This is great. I'd been worrying about my font miles. Locally sourced fonts
are much better, and I feel as though I'm supporting local fontsmiths and
helping preserve traditional skills. Top work.

PR #1493, opened 2018-04-09
A refactor improvement that delegates filtering for serialization dow…

fredkingham:
…n to the individual subrecords

### DIFF
--- a/opal/core/pathway/steps.py
+++ b/opal/core/pathway/steps.py
@@ -16,9 +16,9 @@ def delete_others(data, model, patient=None, episode=None):
     from opal.models import EpisodeSubrecord, PatientSubrecord
 
     if issubclass(model, EpisodeSubrecord):
-        existing = model.objects.filter(episode=episode)
+        existing = model.objects.for_episode(episode=episode)
     elif issubclass(model, PatientSubrecord):
-        existing = model.objects.filter(patient=patient)
+        existing = model.objects.for_patient(patient=patient)
     else:
         err = "Delete others called with {} requires a subrecord"
         raise exceptions.APIError(err.format(model.__name__))
--- a/opal/managers.py
+++ b/opal/managers.py
@@ -3,6 +3,7 @@ Custom managers for query optimisations
 """
 from collections import defaultdict
 import operator
+from opal.core.fields import ForeignKeyOrFreeText
 
 from django.db import models
 from django.db.models import Q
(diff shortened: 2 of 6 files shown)


opal/managers.py
```diff
@@ -68,21 +66,24 @@ def serialised(self, user, episodes,
...
+
         patient_ids = [e.patient_id for e in episodes]
+        patients = Patient.objects.filter(id__in=patient_ids)
         patient_subs = defaultdict(lambda: defaultdict(list))
 
         episode_subs = self.serialised_episode_subrecords(episodes, user)
         for model in patient_subrecords():
             name = model.get_api_name()
-            subrecords = model.objects.filter(patient__in=patient_ids)
+            subrecords = model.objects.for_patients(patients)
```
davidmiller: Doesn't this mean we do an extra query (above) compared to adding prefetch related here ?

davidmiller:
Is this not mostly about speed rather than API ? 
In which case do we have before & after numbers?

opal/core/pathway/steps.py
```diff
@@ -16,9 +16,9 @@ def delete_others(data, model, patient=None, episode=None):
     from opal.models import EpisodeSubrecord, PatientSubrecord
 
     if issubclass(model, EpisodeSubrecord):
-        existing = model.objects.filter(episode=episode)
+        existing = model.objects.for_episode(episode=episode)
```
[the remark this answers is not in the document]
fredkingham: if we did filter rather than for_episode here then we would delete all of the things even if we had puposefully excluded them using the `for_episode` method

fredkingham:
So this is was about API to exclude lab tests that we didn't want serialised (take a look at https://github.com/openhealthcare/opal-lab/pull/45 for how this works). 

While I was working on it, I spent a long time trying to make it so that by wiring in the correct prefetches we could get away without having to have `Episode.objects.serialised` and `Episode.to_dict`. The conclusion sadly was that I could not make it. 

Mostly because `EpisodeSubrecord.objects.filter(episode=episode.id)` was always five times faster than `episode.episodesubrecord_set.all()`

Which mean prefetch related could not give me the speed improvement to make it worth while.

So then I added in the manager methods to give us the hooks that one could use.

Speed times. So on RFH with the latest data...
https://github.com/openhealthcare/opal/tree/380-speed-test is my speed test branch using `./manage.py speed_test` to give me an average serialisation times of the last 100 Episodes on test.

without prefetch with normal method
0.87301473856

with the old method, copying the management command over
0.693269588947

with the new method with prefetch
0.407850282192

Times did vary with repeated runs. These are the slowest times out of all the runs I did, but the new way is always significantly faster.

opal/managers.py
fredkingham: possibly, I thought for_patients, was a nicer api than for_patient_ids

fredkingham:
I think we're killing this, the speed improvement has been merged into opal

PR #1496, opened 2018-04-11
adds a prefetch method that will prefetch many to many and foriegn ke…

### LINKED ISSUE #1495: select_related fk_or_ft_ids
when we serialise an episode we do a lot of queries to get fk_or_ft fields. If we prefetch with a select_related call we _should_ see a big speed improvement

fredkingham:
…ys from Foreign key or Free Text fields, fixes #1495 

don't merge, pending prod comparison data...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@ Print/screen differences are now in opal.css with media tags.
 The loading in of Google Analytics is now deferred to the bottom of the body
 tag to allow the page to load without waiting on analytics scripts to load.
 
+#### Episode.objects.serialised now uses select_related
+ForeignKeyOrFreeText now have their ForeignKey items preselected when
+we use Episode.objects.serialised. This gives a not insignificant speed boost
+(approx 30-40% on some of our real data although obviously this depends on
+your ForeignKeyOrFreeText useage).
+
 ### 0.10.0 (Major Release)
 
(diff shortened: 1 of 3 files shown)


fredkingham:
so we were on
0.776162023544

now we're on
0.411258296967

test script is below
```
"""
Create singletons that may have been dropped
"""
import time

from django.core.management.base import BaseCommand
from django.contrib.auth.models import User

from opal.models import Episode


class Command(BaseCommand):
    def test_run(self, prefetch):
        last_hundred = Episode.objects.count() - 100
        before = time.time()
        Episode.objects.serialised(
            User.objects.first(),
            Episode.objects.filter(id__gte=last_hundred),
            prefetch=prefetch
        )
        return time.time() - before

    def add_arguments(self, parser):
        parser.add_argument(
            '--prefetch',
            action='store_true',
        )

    def handle(self, *args, **options):
        times = []
        for i in xrange(20):
            times.append(self.test_run(options["prefetch"]))

        print "=" * 10
        print "time taken with {} {}".format(
            options["prefetch"],
            sum(times)/100
        )
        print "=" * 10
        return
```

davidmiller:
I think there's a bug in your test command (`times.append(self.test_run(options["prefetch"]))`) but this looks great - running it on a local development application w/out much data we have less pronounced speedup, but still faster:

2.71409776806 down to 2.09484482407

Issue #1497, opened 2018-04-12
FileNotFound error when executing opal startproject myapp


dumodem:
C:\Users\User>opal startproject mynewapp
Bootstrapping your Opal project...
Creating project dir at C:\Users\User\mynewapp
Creating Database
Calling: python mynewapp\manage.py makemigrations mynewapp --traceback
Migrations for 'mynewapp':
  mynewapp\mynewapp\migrations\0001_initial.py:
    - Create model Allergies
    - Create model Demographics
    - Create model Diagnosis
    - Create model Investigation
    - Create model Location
    - Create model PastMedicalHistory
    - Create model PatientConsultation
    - Create model SymptomComplex
    - Create model Treatment
Calling: python mynewapp\manage.py migrate --traceback
Operations to perform:
  Apply all migrations: admin, auth, authtoken, axes, contenttypes, mynewapp, op
al, reversion, sessions, sites
Running migrations:
  Applying contenttypes.0001_initial... OK
  Applying auth.0001_initial... OK
  Applying admin.0001_initial... OK
  Applying admin.0002_logentry_remove_auto_add... OK
  Applying contenttypes.0002_remove_content_type_name... OK
  Applying auth.0002_alter_permission_name_max_length... OK
  Applying auth.0003_alter_user_email_max_length... OK
  Applying auth.0004_alter_user_username_opts... OK
  Applying auth.0005_alter_user_last_login_null... OK
  Applying auth.0006_require_contenttypes_0002... OK
  Applying auth.0007_alter_validators_add_error_messages... OK
  Applying auth.0008_alter_user_username_max_length... OK
  Applying authtoken.0001_initial... OK
  Applying authtoken.0002_auto_20160226_1747... OK
  Applying axes.0001_initial... OK
  Applying axes.0002_auto_20151217_2044... OK
  Applying axes.0003_auto_20160322_0929... OK
  Applying opal.0001_initial... OK
  Applying opal.0002_auto_20150822_0820... OK
  Applying opal.0003_speciality... OK
  Applying opal.0003_auto_20150922_1825... OK
  Applying opal.0004_merge... OK
  Applying opal.0005_team_visibile_in_list... OK
  Applying opal.0006_auto_20151109_1232... OK
  Applying opal.0007_tagging_archived... OK
  Applying opal.0008_auto_20151216_1803... OK
  Applying opal.0009_glossolaliasubscription... OK
  Applying opal.0010_auto_20160229_1211... OK
[...]

PR #1498, opened 2018-04-12
Fail more gracefully in startp[roject|lugin] if git missing

davidmiller:
As reported in #1497 we are currently raising an exception in `startproject` if we can't execute `git` in a subprocess.

This change makes those failures informative message rather than a stack trace.


fredkingham:
works a treat

PR #1499, opened 2018-04-12
As menuitem

davidmiller:
Enable the `x.as_menuitem` pattern for Patient lists and pathways.

Add basic `get_absolute_url` implementation to patient lists and pathways.

Adds a `get()` utility function to look for getters.


opal/core/exceptions.py
```diff
@@ -37,3 +37,7 @@ class InitializationError(Error):
 
 class MissingTemplateError(Error):
     pass
+
+
+class SignatureError(Error):
```
fredkingham: Is this used anywhere?

opal/core/pathway/pathways.py
```diff
@@ -21,14 +22,57 @@ def redirect_url(self, user=None, patient=None, episode=None):
...
+            if isinstance(slugattr, string_types):
+                return slugattr
+        return slugify(klass.__name__)
+
+    @classmethod
+    def get_absolute_url(klass, **kwargs):
+        """
+        Returns a string which is the absolute URL of this Pathway.
+        """
+        return '{0}#/{1}/'.format(reverse('pathway_index'), klass.get_slug())
```
fredkingham: so by default we're saying that a pathway will not receive a patient/episode which I think makes sense.

We have to have one default, and this is the default we would use in a menu item. I think this is correct.

PR #1501, opened 2018-04-14
1489 api router initialization

### LINKED ISSUE #1489: move the api router registration into urls
At the moment we register the plugin apis in 
https://github.com/openhealthcare/opal/blob/v0.10.1/opal/core/api.py#L415

That means that if you want to use any of the ever so handy functions/parent classes such as `patient_from_pk` or `LoginRequiredViewset` you will get a circular import (core api is importing from plugins which are importing from core api).

Move the router registration in to urls to put an end to this tomfoolery.

davidmiller:
I think this is probably better...

Holding off documenting the decorators for now as I suspect we might want to review/refactor if we're going to suggest reuse.

Closes #1489

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,17 @@
 
 #### Removes "episode_history" from episode serialization
 
-Serialised episodes previously contained a "shallow" copy of all other episodes in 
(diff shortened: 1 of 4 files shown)


fredkingham:
excellent

PR #1504, opened 2018-04-14
1502 test options bugfix

### LINKED ISSUE #1502: Combining --test and --coverage options from `opal` doesn't behave as expected
We'd ideally want this to run coverage but with the subset given in `--test` / `-t` In fact, it ignores `-t` as implemented.

Why would you want to do this? 

Well for instance, to check that the coverage you have of the codebase is actually being provided by the test file that has unittests for that module rather than incidentally from tests which aren't actually testing this path.

e.g. if the thing that alerts me to a bug in the implementation of `opal.managers.EpisodeQueryset.serialised` is a test in `opal.tests.test_patient_lists` then that means our tests for `opal.tests.test_managers` aren't great. At that point, looking at test coverage for files when running a subset of test cases is useful.

davidmiller:
If we move the order in which we construct the subprocess string, you get to combine `-t` and `-c`.

closes #1502

### DIFF
--- a/opal/core/test_runner.py
+++ b/opal/core/test_runner.py
@@ -32,21 +32,21 @@ def _run_py_tests(args):
     if _has_file(args.userland_here, 'runtests.py'):
         test_args = ['python', 'runtests.py']
 
-        if args.test:
-            test_args.append(args.test)
-
         if args.coverage:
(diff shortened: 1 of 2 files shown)


davidmiller:
I would be surprised if appveyor didn't go green once we merge in #1505 ...

PR #1507, opened 2018-04-16
1443 can we stop double clicks in the spread sheet view

pacharanero:


pacharanero:
closes https://github.com/openhealthcare/elcid/issues/1443

davidmiller:
As discussed briefly earlier, the alternative to this would be a directive that explicitly debounces clicks...

Similar to https://github.com/angular/angular.js/issues/9826#issuecomment-61654326

```html
<li class="edit_item"
      ng-click="editNamedItem(row, '{{ column.name }}', row.{{column.name}}.length)"
      ng-dblclick="editNamedItem(row, '{{ column.name  }}', row.{{column.name}}.length)"
      debounce-clicks
>
<!-- content -->
  		    </li>
```
Or perhaps even with an optional `debounce-clicks="200"` if we were feeling like it.

The advantage of this is that it's set at the point that we register the click handler rather than in an abstract decoupled location...

pacharanero:
OK sure, seems reasonable

Issue #801, opened 2016-07-28
How would we like to access subclasses of abstract models from pluggable modules?


davidmiller:
(The conceptual equivalent of Django's `get_user_model`).

An alternative might look like this:

``` python
import sys, imp, types

class MagicModule(types.ModuleType):
    def __getattr__(self, attr):
        from opal.core import subrecords
        try:
            return subrecords.get_subrecord_from_model_name(attr)
        except ValueError:
            pass # No subrecord found - use the default response
        return getattr(types.ModuleType, attr)

class ImportMagic(object):
    def find_module(self, fullname, path=None):
        if fullname == 'opal.archetypes':
            return self
        return None

    def load_module(self, name):
        return MagicModule(name)

sys.meta_path.append(ImportMagic())

# In some app: 

from opal.archetypes import Diagnosis
# -> The current subclass of Diagnosis.
```

In practice we might like to do some better error handling here, but this would be a plausible first pass.

davidmiller:
Honestly I still _quite_ want to do this. But closing for now as it's been 18 months. 

Let's revisit when we have a good use case perhaps... 

Fairly sure there will be some.

PR #1512, opened 2018-04-18
Add a custom meta path to allow us to import subrecords by name regardless of their actual location in a file.

davidmiller:
For now, I propose adding this as an undocumented, not officially supported helper / syntax sweetner.

See how we feel about it and then if it seems useful we can document/support it.

refs #801


davidmiller:
Looks like this only works on python < 3.4 

Shelve for the foreseeable

PR #1514, opened 2018-04-20
Only show the parens around DOB if there is a DOB in patient detail h…

davidmiller:
…eader


fredkingham:
So I really want this. Is the below not simpler?

```
<span ng-show="patient.demographics[0].date_of_birth">
 ([[ patient.demographics[0].date_of_birth | shortDate ]])
</span>
```

Or is there a css reason not to do this or something?

davidmiller:
sure - done that way here: https://github.com/openhealthcare/opal/pull/1542

Issue #1513, opened 2018-04-18
If there is only one list, hide the list switcher


fredkingham:
Just running through the tutorial. Add a new app. We get an app with a list switch that goes no where.

Shall we just hide it if there is only a single list?

<img width="327" alt="screen shot 2018-04-18 at 18 21 00" src="https://user-images.githubusercontent.com/2175455/38947665-665c5f1c-4335-11e8-82fe-ebaf1c326616.png">

pacharanero:
@fredkingham IIRC you have an Opal app with one PatientList, would you be able to test whether this is fixed by #1517 ?

fredkingham:
@pacharanero  it was just a new opal app. I just ran through the tutorial prior to the last release. I'm afraid I've nuked the app now, can you create a new app and test it? (Sorry!)

pacharanero:
OK I have checked and can confirm that with one patient list in a new Opal app, the ListSwitcher is suppressed using my local branch of Opal `1477-one-spreadsheet-list-to-rule-them-all` via setup.py develop but not suppressed when using Opal v0.10.0.

pacharanero:
closed by #1517

Issue #1457, opened 2018-03-20
Patient List Groups in the PatientList switcher


davidmiller:
Where we're using `TabbedPatientListGroup` we may well not want each member in the list switcher.

- [ ] Add a property to `opal.core.patient_lists.TabbedPatientListGroup` which promises that, if True, member lists will be removed from the list switcher 
- [ ] Add an `is_valid` check to ensure that if this property is True, we can `get_slug` and that there is a display name
- [ ] Update `opal.views.PatientListTemplateView.get_context_data` such that it does not include lists which are members of a group where this property is True in the `'lists'` variable
- [ ] Consider refactoring this logic out of the view and into `opal.core.patient_lists` as a function 
- [ ] Tests
- [ ] Documentation

This ticket unblocks openhealthcare/elcid#1509

pacharanero:
I've been looking further at this following call/pairing with @davidmiller 

I think the proposed implementation will work adequately, and will implement it as such, but it feels a bit clunky to be using `display_member_lists_in_list_switcher = False` exclued ALL the members from the Patient ListSwitcher, without any granularity.

I was wondering if an eventual API with more flexibility and clarity would be something like this:

```python
member_lists = [
    (WalkinNurseTriage, list_switcher_item=False),
    (WalkinDoctor, list_switcher_item=False),
    (WalkinReview, list_switcher_item=False),
]
display_name = "Walkin" # optional
```

this gives complete flexibility as to which PatientLists are shown in the ListSwitcher, ranging from just `display_name` on its own, all the way up to any combination of the members.

pacharanero:
some initial WIP on this branch: https://github.com/openhealthcare/opal/tree/1457-patient-list-groups-in-the-patientlist-switcher

pacharanero:
There is also some code in elcid-uch which facilitates testing/development of this feature. It is in branch https://github.com/openhealthcare/elcid/tree/1509-given-tabbed-navigation-can-we-make-all-of-opat-one-option-on-teams-dropdown

pacharanero:
Removing the `PatientList` objects from the `list_items` list using a list comprehension and finding them by `display_name` feels really hacky. Is there a better way I'm not seeing?

https://github.com/openhealthcare/opal/blob/fefcea26336a20aa15766e46297b9573da1a782f/opal/views.py#L71-L76

davidmiller:
You need access to the current list group from that method I think.
[...]

PR #1517, opened 2018-04-23
refactor duplicated spreadsheet_list.html 

### LINKED ISSUE #1477: One spreadsheet list to rule them all
We have 2 spreadsheet lists.

https://github.com/openhealthcare/opal/blob/v0.10.0/opal/templates/patient_lists/spreadsheet_list.html
and
https://github.com/openhealthcare/opal/blob/v0.10.0/opal/templates/patient_lists/layouts/spreadsheet_list.html

we would like 1.

Ideally this will have 
- [x] will be the layouts/spreadsheet_list.html (ie we should be able to delete the other one)
- [x] the episode actions block that @pacharanero put into v0.10.0
- [x] make sure we don't lose anything, i.e blocks (inc the episode_actions block introduced by @pacharanero )
- [x] change the patient lists to default to the layouts
- [x] update the change log to state this change

pacharanero:
Closes #1477 

This is also work in progress towards fixing https://github.com/openhealthcare/elcid/issues/1509 via #1457 

Should also close https://github.com/openhealthcare/opal/issues/1513 because there was already a fix in the spreadsheet_view_base.html (which wasn't being used). It is now. I haven't yet tested this but if anyone has an Opal app with just one patient list, a confirmed fix here would enable us to close.


opal/templates/patient_lists/layouts/spreadsheet_list_base.html
```diff
@@ -3,27 +3,109 @@
 
 <div class="container-fluid content-offset">
 
+  {% block heading %}
```
davidmiller: I see what you've done here - this is what is in the current list..

We might prefer to have naming convention consistency with the other two lists though, and use `  {% block list_actions %}`.

davidmiller: (that means deleting the `spreadsheet_list_actions` block as well obv.)

davidmiller:
Overall looks great - just a couple of minor tweaks and then we're good to go!

opal/templates/layouts/left_panel.html
```diff
@@ -1,18 +0,0 @@
-<div class="container-fluid content-offset">
-  {% block heading %}
```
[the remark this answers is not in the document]
pacharanero: Have added to template removal in changelog, and I cannot find any other instances of it being used anywhere else

opal/templates/patient_lists/layouts/spreadsheet_list_base.html
pacharanero: so is `heading` superfluous?
can it be removed,  and change `spreadsheet_list_actions` to `list_actions`?
I'm not 100% clear on the required end state

PR #1521, opened 2018-04-24
1515 serialization refactor

### LINKED ISSUE #1515: Move serialization utilities into one place.
We currently scatter our serialization utilities around the place - `core.views`, `models` both contain useful components we don't use. 

Collect these into `core.seralization` to make reuse easier. 
n.b. the ones in `models` have somehow escaped explicit unittesting - they're only implicitly tested.

davidmiller:
Fixes #1515 

https://github.com/openhealthcare/opal/issues/1515


fredkingham:
Yes please and thank you.

PR #1522, opened 2018-04-24
1520 refactor the switcher dropdown

pacharanero:
* Brings together all processing of the contents of the Switcher into views.py and out of templates.

* Removes `list_slug` which was only being used to suppress the currently viewed list from the dropdown menu - this is now handled in views.py logic

* Removes `num_lists` which was being used in the template to suppress the display of the Switcher in the case when there is only one list. This is also now handled in views.py logic

* Introduces `default_tab` and `display_name` as required properties of a TabbedPatientListGroup. TPLGs are represented on the Switcher by their `display_name` and the `default_tab` is used to generate an initial `slug`. (The plan is that the frontend will cache the last visited tab and this will over-ride the default, but this is not yet implemented)

* Adds documentation for the new properties `default_tab` and `display_name`.

* edited `test_views.py` mainly to remove the test cases which test for `list_slug` and `num_lists`, as these are no longer used. There is still a [failing test](https://github.com/openhealthcare/opal/blob/5ca6d4d143d5f9d7960ee14b01fb14dc208a7d06/opal/tests/test_views.py#L109) which I am unsure how best to fix.

* I have tried to minimise my use of the term PatientList in describing what entities can be part of a Tab Group, and what can be in the switcher. [In the future we may want to be able to include (a subset of?) all `Discoverable` features (and maybe others like Dashboards) in the Switcher and in Tab Groups?]

### DIFF
--- a/opal/templates/patient_lists/partials/list_dropdown_menu.html
+++ b/opal/templates/patient_lists/partials/list_dropdown_menu.html
@@ -1,5 +1,6 @@
 {% load forms %}
-{% if num_lists > 1 %}
+<!-- suppress the Switcher if there are no items in it -->
+<!-- {% if switcher_items != None %}
 <div class="btn-group pull-left" uib-dropdown is-open="status.isopen">
   <button type="button" class="btn btn-secondary" uib-dropdown-toggle ng-disabled="disabled" id="list-dropdown">
     <i class="fa fa-user-md"></i>
@@ -7,15 +8,13 @@
     <i class="fa fa-angle-down"></i>
   </button>
   <ul class="uib-dropdown-menu slides" role="menu" aria-labelledby="list-dropdown">
-    {% for list in lists %}
-    {% if list.get_slug != list_slug %}
+    {% for item in switcher_items %}
     <li>
-      <a href="#/list/{{ list.get_slug }}">
-        {{ list.display_name }}
+      <a href="#/list/{{ item.get_slug }}">
+        {{ item.display_name }}
       </a>
     </li>
-    {% endif %}
     {% endfor %}
   </ul>
 </div>
-{% endif %}
+{% endif %} -->
--- a/opal/templates/patient_lists/partials/list_dropdown_menu.html
+++ b/opal/templates/patient_lists/partials/list_dropdown_menu.html
@@ -1,12 +1,17 @@
 {% load forms %}
+
 <!-- suppress the Switcher if there are no items in it -->
-<!-- {% if switcher_items != None %}
+{% if switcher_items != None %}
 <div class="btn-group pull-left" uib-dropdown is-open="status.isopen">
+
+  <!-- displays the name of the currently viewed item on the Switcher button -->
   <button type="button" class="btn btn-secondary" uib-dropdown-toggle ng-disabled="disabled" id="list-dropdown">
     <i class="fa fa-user-md"></i>
     {{ patient_list.display_name }}
     <i class="fa fa-angle-down"></i>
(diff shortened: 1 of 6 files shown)


opal/views.py
```diff
@@ -50,23 +50,46 @@ def get_context_data(self, **kwargs):
...
-        context['lists'] = list(PatientList.for_user(self.request.user))
-        context['num_lists'] = len(context['lists'])
+        context['switcher_items'] = self.get_switcher_items()
+        context['list_group'] = self.get_list_group_context()
+        context['columns'] = self.get_column_context(**kwargs)
+        return context
 
-        context['list_group'] = None
+    # builds a list of Switcher items filtered for user, tab groups and current item viewed
+    def get_switcher_items(self):
```
davidmiller: If I'm reading the intention correctly, I might spell this something like:
```python
lists = [p for p in PatientList.for_user(self.request.user) if p != self.patient_list]
return {
    TabbedPatientListGroup.for_list(l) if TabbedPatientListGroup.for_list(l) else l for l in lists
}
```
(I've not actually tested that this works BTW)

opal/templates/patient_lists/partials/list_dropdown_menu.html
```diff
@@ -1,21 +1,26 @@
 {% load forms %}
-{% if num_lists > 1 %}
+
+<!-- suppress the Switcher if there are no items in it -->
+{% if switcher_items != None %}
```
davidmiller: Does this work ?

I _think_ that python truthines would work on `{% if switcher_items %}` ([] evaluates to bool False) but it can't see a path for it to be None from the view ?

pacharanero:
Here's the refactor as it currently is. Comments and suggestions appreciated @davidmiller @fredkingham esp. re: testing the changed functionality (which things to test, and approaches for doing so)

As this stands it would close 

Please note that to help you test and evaluate there is a companion branch of elCID-UCH here https://github.com/openhealthcare/elcid/tree/opal-1520-refactor-the-switcher-dropdown which has the new properties on the Walkin and OPAT `TabbedPatientListGroup`s

once complete and merged:
closes https://github.com/openhealthcare/opal/issues/1457
closes https://github.com/openhealthcare/elcid/issues/1509

davidmiller:
If I read this correctly we change the behaviour already, within this one ? e.g. it swaps out 
`[WalkinNurseTriage, WalkinDoctor, WalkinReview, ...]` for `[Walkin,...]` but with the link pointing to the `default_tab` ?

We could do that - there's definitely an argument that it's not worse than the current thing :) 

My thought had been to refactor the templates without changing the behaviour as a first pass, then add a front end service / metadata that we didn't actually call second, then swap the _contents_ of the switcher once that was in  third... you could use default tab as an interim though - thoughts @fredkingham  ?

davidmiller:
Good question re. Tests... I would expect explicit unittests for these methods:

* `TabbedPatientListGroup.get_slug` (to include failure cases as well)
* `get_list_group_context` method
* `get_switcher_items` method
* `get_context_data` return value 

I can imagine you may want to mock `PatientList.for_user` in the tests for some of these so that it becomes easy to hardcode the expected return values 

e.g. you know it's going to be only certain things, and you can look at the 'test' list groups at the top of `opal.tests.test_patient_lists` - easier with a concrete example though !

pacharanero:
> My thought had been to refactor the templates without changing the behaviour as a first pass

Ah. yes you're right that was the plan. Sorry, got carried away and changed behaviour

PR #1530, opened 2018-05-23
bumps compressor version and fixes the pip upgrade command for appveyor

fredkingham:


fredkingham:
Example app with the compressor change http://elcid-something.herokuapp.com/

davidmiller:
I think it needs upgrade docs b/c there's a dependency upgrade

davidmiller:
Needs to be in `setup.py` as well obviously.

doc/docs/reference/upgrading.md
```diff
@@ -3,6 +3,17 @@
 This document provides instructions for specific steps required to upgrading your Opal
 application to a later version where there are extra steps required.
 
+### 0.10.1 -> 0.10.2
+
+The previous Django upgrade stopped compressor minifying files When DEBUG is set to False.
```
davidmiller: I don't think you need to talk about the problem here - this document is 'what do I need to do' e.g. install an updated dependency. 

This text doesn't actually tell me that I need to upgrade something, or that I'll have to update my requirements.txt though - the audience for upgrading.md is someone not following Opal development closely, and potentially requiring detailed instructions to work through the steps required...

doc/docs/reference/upgrading.md
```diff
@@ -38,6 +49,7 @@ you have specified them in for instance, a requirements.txt.
     requests==2.18.4
     django-celery==3.2.2
     celery==3.1.25
+    django-compressor==2.2
```
davidmiller: I see what you're doing here, but this will mean that we have several other versions of `/docs/x/y/upgrading/` which don't have this line?

If we're doing it, we need to change the other branches as well surely?

fredkingham:
re-pointing this as v0.11.0

Issue #1390, opened 2018-01-28
foriegn key and free text fields do not do look ups with case insensitivity please fix.


fredkingham:
Malignant Otitis Externa is the same thing and Necrotising Otitis Externa.

If I save a patient with necrotising otitis externa.
I go to the extract page, and search by all patients with Necrotising Otitis Externa
I should see that there are the same number that I would see if I searched by Malignant Otitis Externa

This did not used to be the case as malignant otitis externa and necrotising otitis externa would both be saved as free text rather than foreign keys.

- [ ] fix synonym look up to be case insensitive.
- [ ] fix lookup list look up to be case insensitive.

fredkingham:
Closing as merged #1616

PR #1534, opened 2018-06-08
foreign key and free text fields are now case insensitive by default,…

fredkingham:
… refs #1390 

Note this doesn't contain a migration to fix this as I think this should probably be done at the application level.


doc/docs/reference/core_fields.md
```diff
@@ -32,3 +32,6 @@ enum('one', '2', 'III')
 
 A field that stores it's value as a generic foreign key to an Opal LookupList
 or as the value in a CharField.
+
+By default this is case insensitive, pass in `case_sensitive=True` to make
```
davidmiller: I'd like a code example below this and more of an explanation of what 'case [in]sensitive' means in this context?

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 ### 0.11.0 (Major Release)
 
+#### Changes all models
```
davidmiller: I think you've committed the changelog with the title stuff in as well - I deleted that blank 'changes all models' heading and moved the title section in a previous PR...

davidmiller:
The reason we don't need a migration is that applications will continue to retrieve FT values which are differently cased versions of lookuplist entries, but stop saving new ones ?

I think we might want a bigger 'warning this upgrade will change the behaviour of your existing models with lookuplists' section in the upgrade guide.

opal/tests/test_core_fields.py
```diff
@@ -1,6 +1,7 @@
 """
 Unittests for the opal.core.fields module
```
davidmiller: You know, I think we may not have any explicit unittests for FT in this file, which I would really expect !

fredkingham: I tested but the True and False cases in the methods


ie This is the case sensitive miss for a lookup list
https://github.com/openhealthcare/opal/pull/1534/files/96b2cfad409cc07583ceecdab4ed63be73371110#diff-7984144a820e90113b1c914866eb396aR117

This is a case sensitive miss for a synonym
https://github.com/openhealthcare/opal/pull/1534/files/96b2cfad409cc07583ceecdab4ed63be73371110#diff-7984144a820e90113b1c914866eb396aR143

fredkingham:
changing this to merge with v0.12.0

PR #1543, opened 2018-06-15
1535 inline date options

fredkingham:
I think I prefer it with the active blue when `today`/`yesterday` is picked, but I'm not sure how I'd deal with 'now' so lets start off with this. The blue can come in as an enhancement if we want it.

screen shot 2018-06-15 at 10 36 57
screen shot 2018-06-15 at 10 36 52
screen shot 2018-06-15 at 10 36 40


fredkingham:
closing in favour of https://github.com/openhealthcare/opal/pull/1544

PR #1545, opened 2018-06-15
exclude subrecords that are "_exclude_from_extract"  from the data dictionary

fredkingham:


fredkingham:
Fixes special casing around episode start and episode end that is no longer relevant.
Makes sure we ignore subrecords that are excluded from the extract in the data dictionary.
Adds additional testing around the data dictionary to make sure our expectations for the fields are correct (ie integration tests)

PR #1548, opened 2018-06-18
Make sure the make docs command works

### LINKED ISSUE #1547: 'make docs' fails
`make docs` looks like it wants to be run in the doc/ directory, however it really doesn't (as then it can't see the makefile, and nobody has time for `make -f ../Makefile docs`.

rossjones:
When running `make docs` the command can't find the mkdocs.yml config
file (it's in doc/).  If you run the command from the doc/ folder, then
make can't see the Makefile.

This PR just makes make make a change in directory before running the
task.

Fixes #1547

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ help:
 
 docs:
 	pip install -r doc/requirements.txt
-	mkdocs serve
+	cd doc && mkdocs serve
 
 release:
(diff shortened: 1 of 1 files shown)


davidmiller:
LGTM - thanks ! 

Just waiting for travis...

PR #1550, opened 2018-06-18
WIP - Allow inference of booleans in enums

### LINKED ISSUE #1540: Form templatetags: raise exceptions when unable to infer an enum
Opal form templatetags attempt to infer enums. They look at Django Choices, Opal Lookuplists or arbitrary JSON passed in the template.

Two enhancements would improve this functionality:

- When passed a Boolean/NullBoolean field, know that the options are e.g. Yes/No/Unknown
- When a widget that requires an enum (Select, Radio) can't find one, raise an exception rather than rendering an empty widget

rossjones:
Where a field is a BooleanField or a NullBooleanField inference should
generate a lookup list with `[Yes, No]` and `[Yes, No, Unknown]` as
options (for radio buttons) respectively.

This will require changes to templates/radio.html so that Unknown does
not use Unknown as a value, instead leaving the value empty so that we
know the intended value is None.

This would be easier if lookuplists were tuples (like choices in Django)
so I'm considering implementing support for lists of tuples (as well as
the current list of strings) for lookuplists - the alternative is too
much logic in the template and special-casing the string 'Unknown'. 
Does this make sense @fredkingham or should I just implement a helper 
for converting Unknown -> "" in this specific case?

## TODO 

- [ ] More tests
- [ ] Ensure value is "" for Unknown
- [ ] Check for i18n 

Will fix #1540


fredkingham:
I think that's a good call. 

It means we'll need changes to do changes to search but maybe supporting choices like Django uses choices is the right thing to do.

fredkingham:
The other side of this argument is that lookup lists will still be either pure text or use each row as the value/text, but I think that's ok.

rossjones:
I was thinking of supporting both options actually, so you can have ["", ""] but for some cases (in particular this one) you can have [("",""), ("","")]

opal/tests/test_templatetags_forms.py
```diff
@@ -449,6 +449,20 @@ def test_radio_infer_lookuplists(self):
...
         self.assertIn('purple', rendered)
 
+    def test_radio_infer_lookuplists_with_boolean(self):
+        template = Template('{% load forms %}{% radio field="HatWearer.wearing_a_hat" %}')
+        rendered = template.render(Context({}))
+        self.assertIn('Yes', rendered)
+        self.assertIn('No', rendered)
+        self.assertNotIn('Unknown', rendered)
+
+    def test_radio_infer_lookuplists_with_quantum_boolean(self):
```
davidmiller: Best test name this week.

opal/tests/models.py
```diff
@@ -37,6 +37,7 @@ class HatWearer(models.EpisodeSubrecord):
     name = dmodels.CharField(max_length=200)
     hats = dmodels.ManyToManyField(Hat, related_name="hat_wearers")
     wearing_a_hat = dmodels.BooleanField(default=True)
+    suits_a_hat = dmodels.NullBooleanField()
```
davidmiller: I think we should pivot to this application.

davidmiller:
ISWYM. It's quite a dirty hack to special case `"Unknown" / None / etc `...

If we're implementing general 'accept tuples' then I reckon you'd want to:

- automatically feed in both halves of fields that have Django choices (I still don't know why you want different values for django choices, but whatever)
- implement for `{% radio|radio_vertical|select %}` (Others?)
- docs for the fact that this is now a thing you can do - presumably the way to roll this on non choices fields is to implement `model.get_field_enum()` ?

And I think we have to update `opal.models.SerialisableFields.build_schema_for_field_name()['enum']` accordingly ? 

And presumably something in `opal/core/search/templates/extract.html` - though @fredkingham may be well placed to advise?

fredkingham:
@davidmiller I think that's a great call, stick the logic in get_field_enum, return a dictionary. 

It will require special casing in extract.html and the two radio button templates.

Yeah that's should work quite elegantly.

rossjones:
For the sake of expedience, and not leaving this PR open for ever, I was investigating the non-elegant special-case the enum in radio.html and it happens that the mix of angular template and django template is pretty horrible if you need to do that sort of thing.  Mostly because the django template runs first, and if the template is bound to an angular [[ ]] then that data isn't available to django :( 

Sorry, just needed to vent a bit :)

fredkingham:
Let it all out!

yeah if you wanted to iterate over something a key values of a js object sometimes but an array other times that get tricky...

I think there's an argument to changing lookup lists to dictionaries but that's way beyond the scope.

Other option would be to change it so that it evaluates the label item (ie not the one on the ng-model) if item == True, item == False, item == None, then put in a label.

I'm not sure that's that nice either...

davidmiller:
Two other options: 

You can look at the field type pre save and convert the string `"unknown"` to `null` in `opal/static/js/opal/services/fields_translater.js` you're after `FieldTranslator.jsToSubrecord`

You can check the field type and the value "Unknown" combination in `opal.models.UpdatesFromDict.Update_from_dict` where we already do type specific work for deserialization of dates/times and coping with many to many fields.

Both of these allow you to continue as if this is any other field with an enum/lookuplist as far as the template is concerned

davidmiller:
Re-pointed at 0.13 so closing this one...

PR #1558, opened 2018-06-22
Lookup list manager

fredkingham:


doc/docs/reference/core_lookuplists.md
```diff
@@ -16,3 +16,26 @@ for l in lookuplists.lookuplists():
...
+
+## {{ LookupListCls }}.objects.search(some_str, case_sensitive=False, contains=False)
+A manager method that takes a string and will search accross the lookup list itself or synonyms. If `contains=True` it will look for partial matches. By default it is case
+insensitive, to change this behaviour query with `case_sensitive=True`
+
+```python
+from opal.core import Antimicrobial
+
+Antimicrobial.objects.search("Difficulty Swallowing")
+## returns the Antimicrobial Dysphagia
```
davidmiller: I'm not medically qualified, but this doesn't look like an antimicrobial to me... can we have an example that is actually the thing that the example is showing ? ;D

doc/docs/reference/core_lookuplists.md
```diff
@@ -16,3 +16,26 @@ for l in lookuplists.lookuplists():
...
+```
+
+## {{ LookupListCls }}.objects.search_many(list_of_strs, case_sensitive=False, contains=False)
+Does the same as the above put takes a list of strings
```
davidmiller: Yeah, but I want to link to this section by `/blah/#this` and get a description of this !
(Though by all means don't do this now if we're going to think about general API changes)

davidmiller:
Is this pushed WIP to have a discussion about or did you decide that finding a `.search(name__icontains='difficulty swallowing)` API wasn't a good idea after all ?

fredkingham:
This is a complete iteration of the original api. I think its up for discussion

I am not sure about `.search(name__icontains='difficulty swallowing)`
Its a wip if we want `name__icontains`. Which I'm not against, I'm just not persuaded.

Arguments for:
* Its more django, you don't need to learn another api.

Arguments against
* There is no django has no understanding of case insensitive `__in` which is going to be our default for search many so we could have search and search many as different functions and then use the __ magic queries seems like a half way ground between the two approaches.. Not sure.

fredkingham:
moving this to shoot for v0.12.0

PR #1559, opened 2018-06-24
Patient demographics property

davidmiller:
Adds a shortcut method for `demographics_set.get()`

I had originally implemented as a property before realising that it would be infuriating to lose the data in this code:

```python
patient.demographics.hospital_number = '1234'
patient.demographics.surname = 'Doe'
patient.demographics.save()
```

Thought about caching the property - but what about race conditions / other updates etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,7 +102,7 @@ a generator function which will yield all subrecord singletons.
 made hospital numbers including `#` or `/` raise an error.
 * Adds the methods `.get_absolute_url()`, `.get_icon()` and `get_display_name()`
 to `opal.core.pathways.Pathway` and `opal.core.patient_lists.PatientList`.
-* Adds a property `.demographics` to `opal.models.Patient` which returns the relevant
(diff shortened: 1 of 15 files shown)


davidmiller:
I think you _could_ have it as a cached property of the Patient object - you just roll back the last commit and add a decorator... 

I'm not strongly convinced either way TBH

How does Django loading foreign keys to properties work with updates to the DB ? 
e.g. `Episode.Patient` - the data is loaded at the moment you first access it, and then never reloaded ? 

We could probably roll back and do that...

davidmiller:
I guess we _could_ do `cached_property(get_demographics, name='demographics;)` and `get_demographics()`

fredkingham:
I've gone the other way, I think this is good, I think as soon as we change it to a cached property we may almost one to change it to a OneToOneField, which we may want to anyway (genuinely don't know, I do think it wouldn't be worth the time and effort).

fredkingham:
closing this to move to v0.12.0

PR #1561, opened 2018-06-25
Lookup list manager

fredkingham:
Adds a lookup list manager for easy seaching of either lookup list names or their synonyms.

This also supports searching for multiple strings within a lookup list


opal/tests/test_managers.py
```diff
@@ -205,3 +210,227 @@ def test_serialised_mine_tag_other_user(self):
...
+            ["hat", "cap"], contains=True
+        )
+        self.assertTrue(
+            hats.filter(id=bowler.id).exists()
+        )
+        self.assertTrue(
+            hats.filter(id=self.hat.id).exists()
+        )
+        self.assertTrue(
+            hats.count(), 2
```
davidmiller: This obviously shouldn't be assert true?

opal/managers.py
```diff
@@ -150,3 +150,71 @@ def serialised_active(self, user, **kw):
...
+            kwargs = {
+                query_arg: some_str,
+                "content_type": content_type
+            }
+            q_objects.append(Q(**kwargs))
+        return Synonym.objects.filter(
+            reduce(operator.or_, q_objects)
+        ).values_list("object_id", flat=True).distinct()
+
+    def get_query_arg(self, contains, case_sensitive):
```
davidmiller: Is this really a public method?

opal/managers.py
```diff
@@ -150,3 +150,71 @@ def serialised_active(self, user, **kw):
...
+        Searches through a lookup list and its synonyms for a value.
+
+        contains runs a contains query rather than an exactly equal
+        query.
+        """
+        return self.search_many(
+            [some_str], contains=contains, case_sensitive=case_sensitive
+        )
+
+    def search_many(self, some_strs, contains=False, case_sensitive=False):
```
davidmiller: Do we need a second method?
Why not `def search(self, *terms,...)` ?

opal/managers.py
```diff
@@ -150,3 +150,71 @@ def serialised_active(self, user, **kw):
...
+    def get_query_arg(self, contains, case_sensitive):
+        if contains and not case_sensitive:
+            return "name__icontains"
+        if contains and case_sensitive:
+            return "name__contains"
+        if not contains and not case_sensitive:
+            return "name__iexact"
+        return "name"
+
+    def search(self, some_str, contains=False, case_sensitive=False):
```
davidmiller: What's the use case for `case_sensitive` ? I can't think of one...

Meanwhile I can't really think why I wouldn't want contains to be True...

opal/managers.py
```diff
@@ -150,3 +150,71 @@ def serialised_active(self, user, **kw):
         episodes = self.filter(**filters)
         as_dict = self.serialised(user, episodes)
         return as_dict
+
+
+class LookupListQueryset(models.QuerySet):
+    def get_content_type(self):
```
davidmiller: I'm not really sure what this is adding?
it's marginally shorter when you call it once below, but at the cost of another method?

opal/tests/test_managers.py
```diff
@@ -205,3 +210,227 @@ def test_serialised_mine_tag_other_user(self):
...
         self.assertEqual(as_dict[0]['tagging'][0], {'id': self.episode.id})
+
+
+class LookupListQuerysetTestCase(OpalTestCase):
+    def setUp(self, *args, **kwargs):
+        super(LookupListQuerysetTestCase, self).setUp(*args, **kwargs)
+        self.hat = self.create_model_and_synonym("Top Hat", "High Hat")
+
+    def create_synonym(self, instance, synonym):
+        ct = ContentType.objects.get_for_model(test_models.Hat)
```
davidmiller: I get why these three methods are here, but it makes it quite hard to read what exactly we're testing in individual test methods below...

opal/tests/test_managers.py
```diff
@@ -205,3 +210,227 @@ def test_serialised_mine_tag_other_user(self):
...
+        return Synonym.objects.create(
+            name=synonym,
+            object_id=instance.id,
+            content_type=ct
+        )
+
+    def create_instance(self, name):
+        return test_models.Hat.objects.create(name=name)
```
davidmiller: Having said I get why the methods are here, this indirection doesn't really seem that useful - it's only called once...

davidmiller:
I think there's thinking about _why_ we're doing this that isn't captured in this PR, and I forget what it is in between looking at it :/

Also, are we sure that putting this on a manager rather than the object itself is the right thing to do?

fredkingham:
I think its a manager method because it deals with many objects. The reasoning was because I had to replicate a bunch of this functionality in search and I think it doesn't really belong in search.

Worth another chat though, lets look at it on Monday.

opal/managers.py
fredkingham: Agreed, this would be nicer, but you can't have *args and then named arguments, 

I've chanced it to take kwargs and extract the values manually.

https://github.com/openhealthcare/opal/pull/1561/commits/7564eb8c3fd396ff8aa1ae4581d9af71bb61bcb9

opal/managers.py
fredkingham: For example in search, we offer, `exact` or `contains`, which is the use case of contains. The case insensitive is for completeness.

I will turn contains on by default

fredkingham:
We don't think we need this at the moment

PR #1563, opened 2018-06-25
1390 free text or fk sensitivity

fredkingham:


davidmiller:
Nothing new into v0.12 - have re-pointed at 0.13

PR #1571, opened 2018-07-13
Terminology

davidmiller:
Before writing docs for this, can we have a review on the implementation....


fredkingham:
doesn't this limit us to a single system/code for a concept?

Slightly wary that https://www.hl7.org/fhir/datatypes.html#codeableconcept explicitly states multiple code/concepts for a `thing`

I'm tempted to add a generic foreign key, even though I'm not a fan. It might be the best way to model this. Thoughts? (Or am I overcomplicating things)

davidmiller:
As a first implementation, I think having only one code & system  for an entry is fine - at least until proved otherwise... b/c complexity, b/c speed...

Things like e.g. translation from LOINC <-> SNOMED are way out of scope for this, which is about giving us the ability to pin individual elements in an Opal referencedata array to points in an e.g. SNOMED graph.

fredkingham:
OK, but we're including a field specifically to keep close to the FHIR spec and then not following the FHIR spec. That's what I'm confused about. Surely we might as well just go the whole hog as I don't think its that much of a work bump. 

I'm happy to whack this on a feature branch and add it in myself if that helps.

davidmiller:
I don't think the general goal for us at this time is really to implement the FHIR spec / data models / resources...

WRT FHIR the goal is to get to a place where it's simple to serialize to a FHIR compliant JSON format rather than to represent information in that format internally.

Things like 'sure have an extra field which will mostly be empty' or 'use this name for the field' I'm very happy to go with what someone else does - if it means increasing complexity, less so...

When push comes to shove, I think we do what makes sense for us and worry about what makes sense for FHIR in `opal-fhir`.

Take out the version field for now is something I could probably get behind honestly...

I'm pushing back a bit because I have no desire to make terminology more complicated than the absolute stupidest we can make it...

I believe that although in SNOMED a codeable concept can occupy multiple places on the graph of all concepts, and thus have multiple codes, as we implement many one dimensional slices of that graph without claiming to be interested in the relation (at least at data entry time) it won't actually come up for us. 

(I could of course entirely be wrong about this ! But it's my current understanding.)

fredkingham:
@davidmiller looks/works great, will merge with docs

fredkingham:
so, lets just go through cases.

We have a lookup list item with a system/code. We remove this. The look up list value removes the code. Is that the behaviour we want? I think it is. We should probably get a unit test for that

So if we put in SNOMED codes in opal, they will be overridden in all old applications that also have the codes in json. Which is fine, just something to be wary of.

Synonyms will be different though. Synonyms will always be created and never deleted. Which is probably wrong? As it breaks the principle of opal can be override by a plugin that can be overridden by an app.

Which would be a different ticket.

Those are some thoughts, maybe we should have a lookup list for overrides code/system with None.

davidmiller:
> We have a lookup list item with a system/code. We remove this. The look up list value removes the code.

I'm not sure I understand what you mean by this?

davidmiller:
> So if we put in SNOMED codes in opal, they will be overridden in all old applications that also have the codes in json. Which is fine, just something to be wary of.

Added not overridden, but sure.
Only if those applications have `opal.referencedata` in installed apps and they haven't forked their referencedata.
Which means they're already accepting Opal being able to change / update that data I don't think this is different.

davidmiller:
> Synonyms will always be created and never deleted. Which is probably wrong? As it breaks the principle of opal can be override by a plugin that can be overridden by an app.

All the load lookuplist stuff is load don't delete though?
If you want to manage your data differently, don't run `load_lookup_lists` run `load_lookup_lists -f your.file.json` or don't have `opal.referencedata` in installed apps?

davidmiller:
> maybe we should have a lookup list for overrides code/system with None.

I'm not sure what this means ? :/

fredkingham:
```
Synonyms will always be created and never deleted. Which is probably wrong? As it breaks the principle of opal can be override by a plugin that can be overridden by an app.
```

I'm still a bit worried about this, but in theory it __shouldn't__ be an issue. 

@davidmiller if you add change log, repoint to 13 I'll merge if that works for you?

PR #1572, opened 2018-07-18
adds a for user method to the menu item, similar to what we do with d…

fredkingham:
…iscoverables

### DIFF
--- a/opal/core/application.py
+++ b/opal/core/application.py
@@ -142,7 +142,7 @@ class OpalApplication(object):
                 items.append(logout)
                 if user.is_staff:
                     items.append(admin)
-        return items
+        return [item for item in items if item.for_user(user)]
 
     @classmethod
     def get_menu(klass, user=None):
--- a/opal/core/plugins.py
+++ b/opal/core/plugins.py
@@ -64,5 +64,5 @@ class OpalPlugin(discoverable.DiscoverableFeature):
         return [j for j in klass.javascripts]
 
     @classmethod
-    def get_menu_items(self, user=None):
-        return [i for i in self.menuitems if i.for_user(user)]
+    def get_menu_items(cls, user=None):
+        return [i for i in cls.menuitems if i.for_user(user)]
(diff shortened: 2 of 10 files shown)


davidmiller:
Looks good, not sure we should have methods of classes in what's going to come out as a <UL > ?

fredkingham:
@davidmiller done

opal/core/plugins.py
```diff
@@ -62,3 +62,7 @@ def get_javascripts(klass):
         Return the javascripts for our plugin
         """
         return [j for j in klass.javascripts]
+
+    @classmethod
+    def get_menu_items(self, user=None):
```
davidmiller: Classmethod so not `self` probably...

opal/core/application.py
```diff
@@ -142,7 +142,7 @@ def get_menu_items(klass, user=None):
                 items.append(logout)
                 if user.is_staff:
                     items.append(admin)
-        return items
+        return [item for item in items if item.for_user(user)]
```
davidmiller: I think we want an explicit test for `Application.get_menu_items` somewhere near here: https://github.com/openhealthcare/opal/blob/v0.12.0/opal/tests/test_core_application.py#L65

opal/core/plugins.py
davidmiller: I think this should probably have some tests of it's own too...

opal/tests/test_core_menus.py
```diff
@@ -107,15 +110,42 @@ def test_sets_items_from_app(self, get_app, plugin_list):
         menu = menus.Menu(user=self.user)
         self.assertEqual(menu_items, menu.items)
 
+    def test_excludes_items_from_app_if_for_user_is_false(
```
davidmiller: Now quite bikeshedding, but I'd like these tests to explicitly call out that what they're really testing is `__init__`

fredkingham: I'm not sure I understand, I don't want to call __init__ because I want is to make sure that this happens on initialisation.

Issue #1574, opened 2018-07-20
Elcid has instances of created/updated by display in the model.


fredkingham:
It's great, lets have that.

fredkingham:
<img width="926" alt="screen shot 2018-07-20 at 15 05 11" src="https://user-images.githubusercontent.com/2175455/43006725-5c7b8b16-8c2e-11e8-8f3d-7ac9728a6e81.png">

davidmiller:
Merged into v0.13.0

PR #1575, opened 2018-07-20
adds created/updated by to the modal foorter refs #1574

fredkingham:


davidmiller:
This looks good - how do I turn it off in an application though?
Can we set a block that I can blank out if I don't want this?

I can presumably easily override part of `form_modal_base.html` for a whole application ?

davidmiller:
Merged into v0.13 instead - closing.

PR #1576, opened 2018-07-28
1573 pathway step save

fredkingham:
the pre_save work exactly as you suggested.

Also previously there was a _bug_ where if you defined the model on a step e.g.

```python
class SomeStep(Step):
   model = MyModel
```
then `SomeStep.model == None`, so I have changed that as well.

There is a pull request ready for rfh when this is merged.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,11 @@
 ### 0.12.0 (Major Release)
 
+#### Pathway Save
+* Pathways steps define what subrecords are saved. By default this is whatever is defined as the `model`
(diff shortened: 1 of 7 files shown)


davidmiller:
Honestly I think the bug with declaring `model = ` on the class would be better split out into it's own PR...

Would likely have merged earlier (at least at the start of these reviews!)

Can we try to remember what the use case for `save_data` is ?
I was struggling...

fredkingham:
This is two years old, and I think we've lost the context.

PR #1579, opened 2018-08-01
changes the episode active flag to depend on the episode category

fredkingham:


davidmiller:
As discussed offline, we're going to leave the flag in the database for now...

PR #1580, opened 2018-08-25
adds the blocks analytics anad javascripts in the base template, allo…

fredkingham:
…wing parents to override

Hopefully not too controversial. Essentially allow overriding of our default analytics/javascripts template behaviour.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,9 @@
 ### 0.12.0 (Major Release)
 
 #### Misc Changes
+* Adds the {% block analytics %} in the base template (opal/templates/base.html) that by default contains the google analytics code.
+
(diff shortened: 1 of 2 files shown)


davidmiller:
I'm not sure I definitely see why you'd want to override these chunks (yet - sure there is a good reason !) ?

fredkingham:
So this all came out of the request to be able to print automagically print an area of the page.

Putting google analytics at the end of a page to get round our proxy problems is fine, but window.print has an implicit call to document.ready so will never complete if the analytics code is present.

We are appending a page onto the main page when we do the window.print. We therefore do not want to call the javascript twice (as its called on load). It also seems like a sensible thing to be able to override.

PR #1584, opened 2018-09-27
Store episode stage in the db

fredkingham:
repointed version of https://github.com/openhealthcare/opal/pull/1486


fredkingham:
Changes to opal/static/js/test/field_translate.service.test.js are spacing and to remove a nested `it` block as this causes issues in new versions of karma

Changes to opal/static/js/test/patient_list.controller.test.js are to remove a nested it block

I also added a test to opal/static/js/test/patient_consultation_record.service.test.js which was missing.

TLDR, the js changes are all test based on prod code based and do not remove any tests.

fredkingham:
@davidmiller this should be good to go I think

opal/models.py
```diff
@@ -1662,6 +1690,35 @@ def explicit_access_only(self):
...
+
+        This is expected to be set by the episode_category
+        which determines what stages are available to an episode.
+    """
+    episode = models.ForeignKey(Episode, null=False)
+    started = models.DateTimeField()
+    stopped = models.DateTimeField(blank=True, null=True)
+    value = models.CharField(max_length=256)
+
+    @transaction.atomic
```
davidmiller: Why does this need to be in a `transaction.atomic` decorator ?

opal/models.py
```diff
@@ -1662,6 +1690,35 @@ def explicit_access_only(self):
...
+    def save(self, *args, **kwargs):
+        """ An episode should not have multiple open stages.
+            raise exceptions and roll back if anyone tries
+            any funny business.
+        """
+        if not self.id:
+            if(self.episode.stage_set.filter(stopped=None).exists()):
+                if not self.stopped:
+                    err = "for episode {}, stage {}. An episode cannot have \
```
davidmiller: Can we look at this error message again? 'For x,y` isn't a regular phrasing, and I think they should start with sentence case...

opal/models.py
```diff
@@ -1662,6 +1690,35 @@ def explicit_access_only(self):
...
+    """
+    episode = models.ForeignKey(Episode, null=False)
+    started = models.DateTimeField()
+    stopped = models.DateTimeField(blank=True, null=True)
+    value = models.CharField(max_length=256)
+
+    @transaction.atomic
+    def save(self, *args, **kwargs):
+        """ An episode should not have multiple open stages.
+            raise exceptions and roll back if anyone tries
```
davidmiller: What do we think we're 'rolling back' ?

opal/models.py
```diff
@@ -1662,6 +1690,35 @@ def explicit_access_only(self):
...
 
 
+class Stage(TrackedModel):
+    """ An episode can have many stages, but should have no more than
+        one with no stop date any time.
+
+        This is expected to be set by the episode_category
+        which determines what stages are available to an episode.
+    """
+    episode = models.ForeignKey(Episode, null=False)
```
davidmiller: Do you need to specify `null=False` I thought that was the default?

opal/core/episodes.py
```diff
@@ -63,20 +65,58 @@ def has_stage(self, stage):
         """
         return stage in self.get_stages()
 
-    def set_stage(self, stage, user, data):
+    @transaction.atomic
```
davidmiller: Can we explain what the desired roll back behaviour / reason to explicitly add a sub-transaction here is?

opal/core/episodes.py
```diff
@@ -63,20 +65,58 @@ def has_stage(self, stage):
...
                 raise ValueError(msg)
-        self.episode.stage = stage
+
+        current_stage = self.episode.stage_set.filter(stopped=None).last()
+
+        # if the stage hasn't changed, don't do anything
+        if current_stage and current_stage.value == stage_value:
+            return
+
+        # some models won't have stages yet, but if an episode category
```
davidmiller: Do you mean "models" here?

opal/core/episodes.py
```diff
@@ -63,20 +65,58 @@ def has_stage(self, stage):
...
+        # has been migrated to use stages,
+        # then it should once an episode has a stage it should
```
davidmiller: Can we re-word this? I'm not quite sure what you mean?

opal/core/episodes.py
```diff
@@ -63,20 +65,58 @@ def has_stage(self, stage):
...
+        # always have a stage.
+
+        if current_stage and not stage_value:
+            raise ValueError(
+                "A stage cannot be removed after one has been set"
```
davidmiller: Can we make the error message explain what has actually happened ? 
e.g. I'm not 'removing' a stage, I'm passing an invalid value, which we know about...

opal/core/episodes.py
```diff
@@ -63,20 +65,58 @@ def has_stage(self, stage):
...
+            )
+
+        # this is the stage for when an episode category
```
davidmiller: What do we mean when we say 'stage' here?

opal/core/episodes.py
```diff
@@ -63,20 +65,58 @@ def has_stage(self, stage):
...
+        if not stage_value:
+            return
+
+        now = timezone.now()
+
+        # close out the old stage and add a new stage
+        if current_stage:
+            current_stage.stopped = now
+            current_stage.updated = now
+            current_stage.updated_by = user
```
davidmiller: We're not doing any checks on User?
That's the only thing I can think of that would raise an error...

opal/core/api.py
```diff
@@ -300,6 +301,7 @@ def list(self, request):
             [e.to_dict(request.user) for e in Episode.objects.all()]
         )
 
+    @transaction.atomic
```
davidmiller: If we're doing this for these... do we not want them for everything? Why just episodes ? Do we really want the whole function wrapped in it? I think we only really want e.g.  from create patient to episode update_from_dict ?

opal/core/episodes.py
davidmiller: I guess I'm wondering if we shouldn't use it as a context manager to wrap those lines to make it more explicit that you want to make sure they're done together....

davidmiller:
As discussed in person, closing for now, opening a ticket which asks for concrete use cases.

PR #1593, opened 2018-10-04
1582 lookuplist synonyms

### LINKED ISSUE #1582: Load lookuplists shouldn't freak out if data doesn't have synonyms
... make the format forgiving!

davidmiller:
Fixes #1582 

Lookuplist data files shouldn't need all those empty synonym lists - just move on if it's not there rather than raising a `KeyError`


fredkingham:
👍

Issue #1578, opened 2018-08-01
The term active should be dependent on the active category


fredkingham:
We have a term active which is defined by whether they have any tags.

However some patients have tags and are still inactive.

Please make it dependent on category so that applications can determine for themselves whether an episode should be active or not.

davidmiller:
Setting it this way is almost certainly often wrong:

```python
        if len(tag_names) and not self.active:
            self.active = True
            self.save()
        elif not len(tag_names) and self.active:
            self.active = False
            self.save()
```

If I was going to pick something right now, I suspect the default would be closer to:

```python
@property
def active(self):
    return self.end == None
```

I can see the value of the indirection to category in letting applications override individual episodes... 

Worth checking to see how active is used perhaps.i

PR #1594, opened 2018-10-04
1578 remove active episode

davidmiller:
Serialising a value for `Patient.to_dict()['active_episode_id']` hasn't made sense since we accepted that there will be multiple concurrent episodes as a common occurrence.

Picking the first episode id from `self.episode_set.filter(active=True).order_by('id').reverse()` is if anything confusing, as it implies that there is some special status to that episode - which there is not.

This PR removes the concept of Patient.active_episode_id and refactors the HospitalNumberCtrl to (the only place that really used it) replicate the behaviour itself.

The behaviour of HospitalNumberCtrl is frankly suspect anyway, but it leaves this PR no more suspect than before it.

Note that we probably want to merge #1590 before this, as Copy To Category also uses active_episode_id.

refs #1578


opal/static/js/opal/controllers/hospital_number.js
```diff
@@ -67,19 +67,37 @@ angular.module('opal.controllers').controller(
...
         };
 
-        $scope.newForPatient = function(patient){
-			if (patient.active_episode_id &&
+      $scope._active_episode_id = function(patient){
+        var active_ids = _.filter(
+          _.values(patient.episodes),
+          function(x){ return x.active == true }
+        );
+        if(active_ids.length > 0){
```
fredkingham: bike shedding but `active_ids.length` is enough

PR #1598, opened 2018-10-04
1597 remove jump to episode from search

fredkingham:
depends on https://github.com/openhealthcare/opal/pull/1594


fredkingham:
Unassigning as I'm going to remove it from the extract as well

davidmiller:
> Unassigning as I'm going to remove it from the extract as well

/nods - do it in both if we're doing it...
I'm going to close this b/c it's going to need re-opening to make the diff look sensible

Issue #1602, opened 2018-10-15
We should not show the `Team` box in the application if there are no teams


fredkingham:
![screen shot 2018-10-15 at 12 45 08](https://user-images.githubusercontent.com/2175455/46949128-39d95f00-d078-11e8-824e-e74edab88f9a.png)

In the tutorial application, we add a patient and there is a weird `Teams` item which does not do anything

innomadic:
I am confused by this Teams box as well (trying to set up for the first time).  How is one supposed to add the teams, if not through this modal dialog box  for editing teams?   As you said, it does nothing.

PR #1603, opened 2018-10-15
only show the teams view if there is at least one team refs #1602

fredkingham:


davidmiller:
I'm closing this as per discussion last week - I think we should use one of those Metadata properties as a proxy and avoid adding new Angular filters as much as we possibly can

PR #1605, opened 2018-10-15
don't use template caching by default refs #1604

### LINKED ISSUE #1604: By default the settings.jinja in opal creates a project with the cached template loader set to True
When you run `opal startproject` you get a project with the cached template loader on. This must me unintuitive for new users

fredkingham:
fixes #1604


davidmiller:
If you want this in 0.12 why not backport https://github.com/openhealthcare/opal/pull/1596/files ?

fredkingham:
I'd forgotten there was a fix in 13, all good I think we don't need this anymore.

PR #1613, opened 2018-10-26
Remove Episode.objects.serialised_active

davidmiller:
Alters the default serialisation of TaggedPatientList serialisation to no longer filter out
'inactive' episodes. Given that 'active' was always true when an episode had a tag, this
was effectivly a no-op anyway unless applications were altering the  for
these patient lists somehow.

refs #1578


fredkingham:
I'm slightly worried, this seems fine but I don't fully understand the repercussions. Given the flag is, I believe not used outside UCH, I'll merge.

PR #1615, opened 2018-10-26
Boolean inference in enums

davidmiller:
refs #1540 

I think we should merge this already as an interim improvement and close the PR at 0.11


fredkingham:
I think we need to rethink this. There are 2 options.

1. Move logic into the field translator and do a hard coded if field type == Boolean or NullBoolean then translate `yes`, `no`, `unknown` accordingly.

2. Use logic similar to what is included here but change it so that we use a dictionary which populates `value`. I'm pro the latter as it doesn't require angular. The problem is any instances where we have hardcoded lookup list properties that currently takes a string that we include as javascript in the template.

Its an interesting one, but lets hold off for the time being.

Issue #1474, opened 2018-03-26
Use of ShortDate filter results in ambiguous dates. Opal default should be DD-Mmm-YYYY. Discuss


pacharanero:
I'm aware that this is essentially reopening https://github.com/openhealthcare/opal/issues/185, but I think the DD/MM date format is very ambiguous and not at all clinically safe.

With a day and month combination that are both <12, the date could be either way round and you have no way of knowing which.

eg 01/09 could mean 1st of September or 9th of January.

For the difference it makes to have the extra few characters I think there's really no good reason not to use a 'fully qualified date' - eg DD/MM/YYYY.

In the NHS/Microsoft Common User Interface, (which are at least an agreed and fairly reasonable set of guidelines, even though never well implemented) within the 'Short Date' standard they advocated the use of DD-Mmm-YYY (where Mmm was a 3-letter non-numeric month representation) which is even less ambiguous. (eg 01-Sep-2018). I think it's quite a nice way to show a date and it scans quickly with the eyes.

The MSCUI site is now dead, but there's a number of (achingly, infuriatingly slow) snapshots of it on archive.org: https://web.archive.org/web/20080612111240/http://www.mscui.net:80/Samples/DateTime/DateTimeSample.aspx The older ones seem more complete and faster because they're less JaVa5cR1PtY

davidmiller:
> very ambiguous and not at all clinically safe.

I'm not sure language like 'not at all clinically safe' is _particularly_ useful... doubly so when you're obviously aware that this has been explicitly reviewed by multiple clinicians and they've found it acceptably safe! (In large part because it replicated the documentation of dates of things that they were currently using clinical practice, with our without our applications IIRC)

I've also rarely found argument from CUI particularly convincing :)

_Anyhow_: focussing on the material content, I take the point about ambiguity. 

Worth remembering the context for the original decision.

Around 30 Aug 2013 applications were very much patient lists rather than patient records.
Almost all user interaction happened in what we'd now think of as a spreadsheet list.
That puts horizontal space at a premium makes the gain/loss of a few characters important.

For some time, there's been the option to replace `|shortDate` with `|momentDateFormat:'DD-MMM-YYYY'` which allows individual applications to have whatever date format they please. Although it does of course require you to override base templates in [quite a number of places](https://github.com/openhealthcare/opal/search?l=HTML&p=1&q=shortDate&type=).

[...]

PR #1623, opened 2018-10-29
1474 swap default date display

davidmiller:
This is item 2/3 in the plan outlined in #1474 

Worth noting that we now have to include `opal.config` in app dependencies to have displayDate render.

As a separate ticket I'm going to suggest adding this to `var implicit_dependencies` in opal/utils.js and removing these explicit includes.


fredkingham:
Please can we update the doc string as I get easily confused. Code wise, looks great, merge when ready.

davidmiller:
Merged on the commandline, closing this

PR #1625, opened 2018-10-30
Fixup: Update markdown table syntax to work with package versions ins…

davidmiller:
…talled when one runs make docs

This should fix the rendering issues.


fredkingham:
👍

PR #1638, opened 2018-11-01
1474 short date filter warnings

### LINKED ISSUE #1474: Use of ShortDate filter results in ambiguous dates. Opal default should be DD-Mmm-YYYY. Discuss

davidmiller:
This is the final PR which closes #1474 

We emit console warnings when short[Date|Time](Time) are used.

The projected removal version is 0.14.0

### DIFF
--- a/opal/scaffolding/record_templates/record_display.jinja2
+++ b/opal/scaffolding/record_templates/record_display.jinja2
@@ -7,7 +7,7 @@
    {% elif f['type'] == 'date_time' %}
     [[ item.{{ f['name'] }} | displayDateTime ]]
    {% elif f['type'] == 'time' %}
-    [[ item.{{ f['name'] }} | hhmm ]]
+    [[ item.{{ f['name'] }} | shortTime ]]
    {% else %}
     [[ item.{{ f['name'] }} ]]
    {% endif %}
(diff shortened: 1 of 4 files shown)


fredkingham:
@davidmiller can we remove the `hhmm` work from this PR. As discuss if we are introducing `displayDate` we should rename `hhmm` to `displayTime`

PR #1651, opened 2018-11-12
Add additional documentation for Episode Category / Patient detail template inheritance.

### LINKED ISSUE #1428: documentation around template inheritance is thin
Generally the documentation is pretty good, especially for the Django bits of Opal. One area that's not very well covered (as far as I can find) is the chain of inheritance/overriding that results in the creation of some of the templates.

For example, I'm trying to create a detail view for an episode category, but am getting a demographic sidebar and episode list from seemingly nowhere. I've had some of it explained by @davidmiller now, but this is an area where the docs could be much clearer.

davidmiller:
Although the ticket talks about template inheritance in general, this PR adds just docs for the part mentioned.

We could definitely use more x-linkage in the docs, and short term happy to add any clarifying sections for any specific things pointed out.

Medium term, completing the theming tutorial (https://opal.openhealthcare.org.uk/docs/tutorials/theming_tutorial/) and / or drastically reducing the role of Angular makes the 'template inheritance' thing go away by reducing the point of ambiguity over 'what is rendering this bit of screen'.

I think this closes that ticket at least as much as we're reasonably likely to write any extensive template inheritance specific docs (not at all likely) in the short term.

closes #1428

### DIFF
--- a/doc/docs/guides/episodes.md
+++ b/doc/docs/guides/episodes.md
@@ -43,6 +43,12 @@ print episode.category.detail_template
 # detail/inpatient.html
 ```
 
+The Episode Category template does not comprise the entire
+[Patient detail view](/patient_detail_views.md). This is made of multiple episodes
+and by default will display some basic demographic details as well as other episodes.
+More detail on customising the rest of the detail tempalte is found in the detail view
+[Template selection docs](/patient_detail_views.md#template-selection).
+
 ## Default Category
 
 The default category of episodes in an Opal application is set by the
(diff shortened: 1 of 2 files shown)


fredkingham:
so... The links don't work for me

screen shot 2018-11-12 at 22 45 25


I haven't __just__ fixed them, because I don't really understand why they don't work but the links in for example datamodel.md that point to `[Archetypes](archetypes.md)` do work...

fredkingham:
created https://github.com/openhealthcare/opal/pull/1651 to turn them into relative paths which we do use elsewhere

PR #1653, opened 2018-11-12
change links to be relative paths rather than point to the .md

fredkingham:


fredkingham:
refs https://github.com/openhealthcare/opal/pull/1651

davidmiller:
How are you running the documentation, and with what versions? (I changed them because I was experiencing that behaviour) 

Is it `make docs` ?

davidmiller:
After looking together, closing this - we should go via .md links where possible

PR #1735, opened 2019-06-13
Adds serialization methods for datetime/time/dates

fredkingham:


davidmiller:
You definitely want three single line methods instead of making public something that figures it out on your behalf ?

fredkingham:
_temporal_thing_to_string isn't public. 

We could make it public but when I'm thinking how do I serialize a date, I think serialize_date, especially as we also have deserialize_date as methods.

PR #1764, opened 2019-08-19
adds a basic add patient pathway to the scaffold for when you create …

fredkingham:
…a new app


fredkingham:
Just adds a demographics add patient. The save button is in a bit of weird place.. But that's not a bug of this, its a pathways bug. I'll take a look in a separate PR

Screen Shot 2019-08-19 at 14 44 26

PR #1797, opened 2019-09-11
Adds an age property to the demographics model

fredkingham:


fredkingham:
I'm going to change this to how we're doing it at rbhl which doesn't require date util. 

Also I want a `age_at` method as well as an `age` property

PR #1812, opened 2019-11-29
show the patient name, identifier in the admin, for episodes it's als…

fredkingham:
…o useful to see the category name

### DIFF
--- a/opal/admin.py
+++ b/opal/admin.py
@@ -83,11 +83,13 @@ class MyAdmin(VersionAdmin):
 
 class EpisodeAdmin(VersionAdmin):
     list_display = [
-        'patient',
-        'active',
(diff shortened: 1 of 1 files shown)


opal/admin.py
```diff
@@ -111,14 +119,20 @@ def view_on_site(self, obj):
...
     ]
 
+    def name(self, obj):
+        return obj.demographics().name
+
+    def identifier(self, obj):
+        return obj.demographics().hospital_number
+
     def patient_detail_url(self, obj):
         return "/#/patient/{0}".format(obj.id)
```
davidmiller: This is probably wrong if we're looking at this now... 
You can do reverse / get absolute url I think

PR #1817, opened 2019-12-06
Allow queries values of false or 0 in the extract

fredkingham:


fredkingham:
> Changelog note about bugfix?

done.

PR #1844, opened 2020-04-06
Allows the date file to be required, either by passing in the require…

fredkingham:
…d field or inferred from the database field


fredkingham:
@davidmiller 👍 done.

PR #1848, opened 2020-04-14
Pathway was not adding the patient to the api call if episode was not…

fredkingham:
… present. Note this will still not add the patient to the pathway to dict.

So this will save to the correct patient but will not add serialised patient subrecords to the front end context, so for example demographics would appear empty.


opal/core/pathway/static/js/test/pathway_loader.service.test.js
```diff
@@ -26,6 +26,14 @@ describe('pathwayLoader', function() {
     expect(!!result.then).toBe(true);
   });
 
+  it('should ad that patient id if the episode id is not provided', function(){
```
davidmiller, line 29: ad that ?

PR #1857, opened 2020-04-23
Scaffold inpatient

fredkingham:


davidmiller:
I want to do https://github.com/openhealthcare/opal/pull/1932 instead ...

PR #1884, opened 2020-07-21
Static simple search

fredkingham:


fredkingham:
The view logic is complicated by 2 things...

#### 1. Why the complicated paginator logics 
We duplicate the logic that was in the paginator which is slightly different from the django default logic. I think its pretty good tbh. I am not sure out logic of when we should show left/right arrows is correct, and I'm not sure if we deal with different screen sizes that well, but this just makes sure it works like it used to.

#### 2. Why is this not a simple list view with get_queryset overridden 
In my infinite wisdom when I wrote the creation of patient summaries, it aggregates them from episodes. So this does Patient.objects.search which returns patients, get's their episodes and then builds the patient summaries from this, it would simplify everything alot if they just took patient as an argument.

The main reason I did this was for filtering out episodes if they weren't visible to the user. I wonder if this could be simplified but this PR is not the place for that.

opal/core/search/views.py
```diff
@@ -242,3 +214,63 @@ def get(self, *args, **kwargs):
...
+        episode_ids = []
+
+        # If a patient has multiple episodes in the list
+        # ignore the additional episodes
+        for episode in episodes:
+            if episode.patient_id in patient_ids:
+                continue
+            episode_ids.append(episode.id)
+            patient_ids.add(episode.patient_id)
+        return models.Episode.objects.filter(id__in=episode_ids)
```
davidmiller, line 247: Why are we returning episodes not patients?

PR #1888, opened 2020-08-03
Removes the notion of subtags as they increase code complexity and ar…

fredkingham:
…e not required


doc/docs/reference/upgrading.md
```diff
@@ -3,6 +3,8 @@
 This document provides instructions for specific steps required to upgrading your Opal
 application to a later version where there are extra steps required.
 
+#### v0.18.3 -> v0.18.4
+PatientList.subtag is no longer a thing. This means adding an episode with a subtag will no longer add it with a tag.
```
davidmiller, line 7: Can we be less flippant in upgrade guides about breaking changes...
Is there any advice about how you would upgrade?
At a guess it's to add a tagging post save hook if you desperately want implicit secondary tags?

davidmiller:
Doesn't this change also alter the allowed characters in tags e.g. dashes - shouldn't that be written in changelogs somewhere? 

Why are we doing it ?

PR #1889, opened 2020-08-03
Tag hierachy was removed from the backend in opal 0.5.* so its probab…

fredkingham:
…ly time to remove it from the front end


davidmiller:
@fredkingham JS tests fail...

Issue #1019, opened 2017-03-09
if an input has choices, have a radio field by default


fredkingham:
at the moment if we scaffold an app with a charfield with choices field it will put in an input field, lets use a radio field

C09:
Hi, 
I wanted to contribute. Is this still an issue.

fredkingham:
Hi Chandni

It hasn't and a contribution would be great.

Some background that may help... 
(this may be obvious, if so sorry!)

In terms of fields with options 4 template tags exist `input`, `select`, `radio` and `radio_vertical` .

Input renders an input field that is free text but with autocomplete options, select is an html select tag and radio/radio_vertical are radio button fields.

When you run `python manage.py scaffold`. `opal.core.scaffold` renders `opal/core/scaffolding/record_templates/record_form.jinja2`

This looks at the schema for the field (the schema is a serialized version of the model structure) and if its a charfield (ie type string in the schema) it renders an `input` template tag, with the autocomplete options.

However the most common case we have found that is if you are rendering a charfield with choices is that you want a radio field.

In the schema the choices are serialized in the `enum` field of the schema.

The solution is that in the `record_form.jinja2` check that a field is a string field and check to see if the `enum` field is populated and if so to use a radio field.

fredkingham:
also https://github.com/openhealthcare/opal/blob/v0.18.4/CONTRIBUTING.md if you haven't seen it

C09:
Hi Fred,

Thank you for replying, I was listening to your interview on the Talk
Python podcast yesterday. Great journey.

I started with installing Opal using pip and playing around with the
application. Next I will fork the repo and create a local environment in my
laptop and start working on your suggestions.

Regards,
Chandni

PR #1890, opened 2020-08-05
scaffolding for string datatype with choices

C09:
#1019 

Allows scaffolding an app with a charfield with choices field


fredkingham:
Looks good! Obligatory _but_ , can we add some tests like https://github.com/openhealthcare/opal/blob/v0.18.4/opal/tests/test_scaffold.py#L631

C09:
yes sure, thanks.

PR #1910, opened 2020-10-07
Add translation tags to sign in page

davidmiller:


fredkingham:
Now merged into the parent change https://github.com/openhealthcare/opal/pull/1911 so closing this

PR #1921, opened 2020-11-01
Django bump

fredkingham:
bumps to django 2.2

Changes from 2.0 -> 2.1 

* auth.views are now class based, so login/change password etc needed to be updated.
* NullBooleanField now inherits from BooleanField so that needed a tweak. (In fact its recommended we use BooleanField(null=True) going forward, they are going to deprecate NullBooleanField)
* {% include 'does not exist' %} now raises an exception so a lot of the tests failed.

Changes from 2.1 -> 2.2
* admin now requires messages

example app
https://opal-app-with-django22.herokuapp.com/#/patient/1/1


opal/scaffolding/plugin_scaffold/runtests.py.jinja2
```diff
@@ -34,6 +34,7 @@ settings.configure(DEBUG=True,
                                    'django.contrib.contenttypes',
                                    'django.contrib.sessions',
                                    'django.contrib.staticfiles',
+                                   'django.contrib.messages',
```
davidmiller, line 37: Don't we have to do this step in all applications & thus should be in the upgrading guide?

fredkingham: so the change from django is that admin now requires messages, so run tests required messages.

The application already scaffolds a settings.py that includes messages so you should not need to add it to the upgrading guide because its already there.

We can do if you think that some people may have removed it manually and now need to add it again?

davidmiller (approved):

PR #1940, opened 2020-11-27
Drop support for python 3.5 from travis and update the upgrading and …

fredkingham:
…change log


fredkingham:
closing because we should fix celery when we do this as our celery libs at present do not officially support python 3.6*

PR #2001, opened 2022-06-08
Stop extract js error

fredkingham:
when you load up the extract screen, there is an error in the terminal, the page works fine but `debugger` breaks if its not in a function.

(also best not to have errors)


fredkingham:
This has already been merged on a different branch

PR #2002, opened 2022-06-08
Exclude many to o from extract choices

fredkingham:


fredkingham:
This has already been merged on another branch

PR #2024, opened 2022-11-29
pin import_lib_metadata

fredkingham:
There is an issue with celery and python 3.7 as described here https://github.com/python/importlib_metadata/issues/411

This resolves that issue.


davidmiller (approved):
passes checks...

PR #2025, opened 2023-02-27
163 disable search while searching

fredkingham:
While the patient is searching for a name/MRN we disable the search button.

Screenshot 2023-02-27 at 18 10 38

### DIFF
--- a/opal/core/search/templates/search.html
+++ b/opal/core/search/templates/search.html
@@ -18,6 +18,7 @@
                      slash-key-focus="!state || state==='normal'"
                      ng-blur="enableShortcuts()"
                      ng-focus="disableShortcuts()"
+                     {# Reenables the search button/form after the input box is changed #}
                      ng-change="searching=false"
                      placeholder="Name or Hospital Number"
                      ng-model="query.searchTerm"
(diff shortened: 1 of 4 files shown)


fredkingham:
The failure is solved by https://github.com/openhealthcare/opal/pull/2024

davidmiller:
You sure the imports PR is pointing at a branch that will affect this?

opal/core/search/templates/search.html
```diff
@@ -10,14 +10,15 @@ <h2>
...
-      <form ng-submit="search()" class="form form-horizontal">
+      <form ng-submit="!searching && search()" class="form form-horizontal">
         <div class="row content-offset">
             <div class="col-sm-4 col-sm-push-4">
               <input type="text" autofocus
                      name="name"
                      slash-key-focus="!state || state==='normal'"
                      ng-blur="enableShortcuts()"
                      ng-focus="disableShortcuts()"
+                     ng-change="searching=false"
```
davidmiller, line 22: I think I want a comment here to highlight what this does - e.g. re-enables the search button if you change the search string.
I don't 100% know how to comment syntax inside indented single html element that won't break rendering/parsing

davidmiller (approved):
Observing users, only really seems to matter when a server is slow, but useful then...

PR #2026, opened 2023-05-18
Bumps the mkdocs opal version

fredkingham:
Bumps the mkdocs version to 0.24.0


davidmiller:
Can we confirm that the documentation definitely still compiles with the new version

davidmiller (approved):

PR #2027, opened 2023-05-18
Fix a bugin create_random_data

fredkingham:
It was possible for the date_generator to error with "ValueError: empty range for randrange()". For a leap year the chance was approximately 1 in 12*88 although this changes with date. This happened when the start date integer entered into random.randint was after the end date integer.

This change makes that chance impossible.


davidmiller (approved):
Honestly, I'm not sure how much we use this generator?

